PR #70, opened 2026-07-12
Add arkitscenes-download: ARKitScenes downloader + layered Rerun ingestion

pablo-rerun:
## What it does

Downloads [ARKitScenes](https://github.com/apple/ARKitScenes) (5,071 indoor iPhone/iPad captures) and ingests each sequence into **seven small rrds sharing one `recording_id`**, so the catalog stacks them as layers and any aspect can be regenerated without re-transcoding video:

| layer | contents |
|---|---|
| `base` | recording properties (clock offsets, orientation, pose provenance) |
| `calibration` | 60 Hz `world_T_rig`, pinholes, 8-coefficient lens distortion (simplecv components), stereo extrinsics |
| `video_wide` / `video_ultrawide` | AV1 `VideoStream`, native resolution/framerate |
| `depth` | ARKit + laser-GT depth, confidence |
| `imu` | 100 Hz accel / gyro / attitude |
| `gt` | annotated mesh + oriented 3D boxes |

Most of this is decoded from the mov's `mebx` metadata streams (NSKeyedArchiver parsing): true 60 Hz camera poses, raw IMU, per-frame distortion polynomials, wide↔ultrawide extrinsics, per-camera clock offsets. Orientation is measured from gravity — the dataset's `sky_direction` label is wrong for ~60% of sequences.

## Try it (5-sequence sample, ~7 GB)

```bash
pixi run -e arkitscenes-download arkitscenes-download-sample
pixi run -e arkitscenes-download arkitscenes-download-ingest
pixi run -e arkitscenes-download arkitscenes-download-serve      # terminal 1
pixi run -e arkitscenes-download arkitscenes-download-register   # terminal 2
pixi run -e arkitscenes-download arkitscenes-download-view
```

Scaling up: `arkitscenes-download-pipeline` runs the resumable chunked driver (download → ingest → ship → sha256-verify → register → clean staging) with local-dir or `user@host:/abs/path` destinations (argv-only tar-over-ssh transport, no shell interpolation). This pipeline has shipped 400+ production sequences (~132 GiB of rrds).

## Port notes

- Monorepo runs python 3.12 (standalone used 3.14): `threading.Lock | None` annotations became `_thread.LockType | None` — `threading.Lock` is a factory function until 3.13 and beartype rejects it in unions.
- A Draco point-cloud compression prototype (`points_compress.py`) was deliberately left behind — laser point clouds are excluded from the pipeline.
- Two tests skip when the sample mov/traj fixtures aren't downloaded.

## Heads-up: monorepo `typecheck` task currently hangs (pre-existing)

While gating this PR I found `pixi run typecheck` spins pyrefly (1.1.0 **and** 1.1.1) in an infinite solver loop on **any file that uses numpy** — including existing packages (reproduced on `simplecv`). The trigger is the `typings/pyrefly/tensor_shapes/fixtures` entry in `pyrefly.toml`'s search-path; removing that one line makes everything check in seconds. It reproduces with this PR's changes fully reverted, so it's not introduced here. This package was gated with the fixtures line removed: **0 errors** (plus ruff, vulture, and 30 tests green).


packages/arkitscenes-download/arkitscenes_download/ingest/blueprint.py
```diff
@@ -0,0 +1,33 @@
...
+        rrb.Vertical(
+            rrb.Horizontal(
+                rrb.Spatial3DView(name="world", origin=f"/{WORLD}"),
+                rrb.Vertical(
+                    rrb.Spatial2DView(name="wide", origin=PINHOLE_WIDE, contents=["$origin/video", f"/{WORLD}/gt/boxes/**"]),
+                    rrb.Spatial2DView(name="ultrawide", origin=PINHOLE_ULTRAWIDE, contents=["$origin/video", f"/{WORLD}/gt/boxes/**"]),
+                ),
+                rrb.Vertical(
+                    rrb.Tabs(
+                        rrb.Spatial2DView(name="depth ARKit", origin=PINHOLE_WIDE_LOWRES, contents=["$origin/depth"]),
```
pablovela5620, line 36: lets make it so that confidence and depth are actually a rrb.vertical so that we can see both at once

pablo-rerun:
Done in 20b960d's follow-up — depth (ARKit/GT tabs) and confidence are now stacked in a `rrb.Vertical`. Also re-registered the updated default blueprint on the local catalog; reopen any sequence to see it.

pablo-rerun:
Three UX rounds since the last comment, all verified pixel-level in the viewer:

1. **Timelines start at 0** — every stream is rebased to a per-sequence epoch, recorded in the properties (`uptime_epoch_seconds`).
2. **No blank first frames** — the epoch is the first instant *both* cameras have a frame: the earlier camera's leading frames are trimmed at transcode time (encoder restarts on a keyframe), all other streams drop pre-epoch samples, and each stream's first sample snaps to t=0. Wide, ultrawide, depth, and confidence are all populated the moment a recording opens.
3. **Blueprint polish** — same layout for both orientations with wider 3D column for portrait sequences (embedded per-sequence; confirmed the embedded blueprint overrides the dataset default when opening a catalog segment), ±5s cursor-relative window on the IMU plots (`TimeAxis(view_range=…)` — note `time_ranges`/`VisibleTimeRange` alone does not constrain the axis), and a 0.25 auto-orbit on the world view.

PR #78, opened 2026-07-22
Eliminate no-build-isolation source builds: prebuilt gsplat, vendored+consolidated MoGe/dinov2, static-metadata sam2

pablo-rerun:
> **Merge order:** merge #77 first (this branch stacks on it; GitHub retargets to main when its branch is deleted). Main (incl. the Rerun 0.35 migration, #81) is merged in and re-gated: parity still bit-exact, suites green.

## What

Scoped strictly to the former `[pypi-options] no-build-isolation` offenders (`moge`, `gsplat`, `dpvo`, `sam2`) — and then a deep simplification of the code that replaced them:

| Package | Before | After |
|---|---|---|
| `gsplat` | PyPI universal wheel that **JIT-compiles CUDA on first use** | prebuilt `ai-demos` conda pkg (`py312_torch212_cuda130`, archs `8.6;12.0+PTX`) |
| `moge` | unpinned git dep dragging two more unpackaged git deps (`utils3d`, `pipeline`) | **vendored inference subset** in `monopriors/third_party/moge` — the three deps vanish from resolution entirely |
| `rtmlib` | *(rescoped)* | **unchanged from main** — it was never a no-build-isolation offender (static fork metadata, empty requirements, already worked on aarch64), so the earlier inline-conda conversion was reverted byte-identical |
| `sam2-streaming` | dynamic `setup.py` importing torch | static `[project]` metadata; CUDA ext opt-in, default off |
| `dpvo` | listed in no-build-isolation | delisted — metadata was already static |

`[pypi-options] no-build-isolation` is gone entirely. Also carried: pixi ≥0.73 named CUDA workspace platforms (stacked from #77).

## The vendoring, and what happened to it

Only `monopriors` imports moge (four `MoGeModel` imports), so the inference subset was vendored function-level into `monopriors/third_party/` — then treated as **owned code** and aggressively simplified:

- **dinov2 consolidated**: moge's copy and the depth-anything-v2 copy (which video-depth-anything already shared) merged into one `third_party/dinov2` — **2,436 → 500 lines**. The two forks differ in live math (moge bakes SDPA into attention; pos-embed interpolation differs ~3.5e-6), so the shared ViT keeps both behind explicit constructor knobs, factory-defaulted per consumer. promptda's torchhub snapshot and g3t's custom variant deliberately left alone.
- **Dead code deleted** (grep-evidenced): xformers/nested-tensor/ViT-G/register/chunked/masked/onnx paths, hub weight downloading, sdpa-wrapper + gradient-checkpointing machinery, the whole `utils3d` shim (its 2 live functions moved into `geometry_torch`), and upstream's latently-broken `refine_depth_with_normal` (calls a function that doesn't exist at moge's own pinned utils3d rev).
- **Loaders hardened**: `strict=True` with unknown-config-key rejection; all five checkpoints revision-pinned to immutable HF shas.
- **Conventions + tooling**: moge/dinov2 now under ruff (148→0 diagnostics), vulture, beartype, jaxtyping/PEP 526 throughout; `Any` eliminated from `monopriors/models` (generic predictor bases + protocols for hub-loaded models); `MogeV1Predictor` → `MoGeV1Predictor` / `moge_v1.py` so the v1/v2 split is explicit.
- The calibrated-FoV inference path (`infer(fov_x=...)`) is kept, reimplemented as one solver instead of two near-duplicates, and now has synthetic round-trip tests.

## Verification — bit-exact, gated at every step

Reference outputs were captured **before** vendoring (upstream moge via the original deps) and every stage re-compared against them:

| Gate | Result |
|---|---|
| moge rrd parity: 2 images × 3 checkpoints (v1, v2-metric, v2-normal) × all raw heads | **48 columns bit-identical, atol=0.0** after every stage (determinism proven first: two reference runs identical) |
| DepthAnythingV2 end-to-end (real vits weights) + unit dinov2 forward | **14 columns bit-identical, atol=0.0** |
| VideoDepthAnything end-to-end (8-frame clip, real weights, pre-consolidation tree vs HEAD) | **24 columns bit-identical, atol=0.0** |
| dinov2 forward goldens (both fork behaviors, checked in as npz) | permanent test, bit-exact |
| monoprior suite (hypothesis property tests, strict-loader tests, goldens) under beartype | 47 passed; lint/pyrefly/vulture clean |
| posekit suite (rtmlib-revert acceptance) | 18 passed (matches the original benchmark) |
| Whole-workspace relock + frozen installs (monoprior, mv-api, posekit, simplecv-catalog) on dl-server | ✓ |
| `pixi lock --check` on pablos-mac-mini (post-rescope, rtmlib back to git-pypi) | ✓ |

Lock state: 0 conda source packages, 0 moge/utils3d/pipeline refs, rtmlib diff vs main = zero lines.

## Notes

- gsplat recipe lives in the ai-demos repo; the sam2 `_C` finding (extension already broken in prod, fallback now explicit) is unchanged from the original PR description.
- Follow-ups deliberately not in this PR: merge from main (Rerun 0.35), promptda's 6,098-line torchhub snapshot getting the same treatment, torch-native focal solver (would change numerics).


pablo-rerun:
Verification addendum: `mamma-dev` test suite against this branch — **14 passed, 1 skipped** (4.11s, dl-server). Both sam2 consumers (posekit: 18 passed, mamma) now validated on the static-metadata sam2 with `_C` off.

PR #80, opened 2026-07-24
ARKitScenes full-corpus pipeline: Modal conversion, gt schema-width fix, hardened catalog registration

pablo-rerun:
## What

The full ARKitScenes corpus run (5,047 attempted → 5,015 published) end to end: Modal conversion fan-out, the gt registration fix, hardened catalog tooling, an abc-130k-style simplification of the local flow, and the runbook capturing every lesson.

### Modal conversion pipeline (`modal_jobs/`)
- Pixi-inside-the-image: byte-identical env to local (conda-forge ffmpeg incl. `av1_nvenc`), build-time encoder asserts, ephemeral HF token.
- 32 L4 converters → staging Volume → **one** `drain_to_hf` with `upload_large_folder`. HF hard-throttles repo commits (~6/min observed); pointing a worker fleet at an HF repo burned ~$45 in 429 backoff before the rearchitect.
- `transfer_to_s3`: Modal OIDC creds, one-call discovery, 16 workers (the xet read-token endpoint 429s at ~28 req/s).
- Full run: ~$265, ~5 h wall-clock vs ~70 h projected locally.

### gt schema-width fix (`ingest/gt.py`) — the big one
Boxes were logged at `/world/gt/boxes/box-<uid>-<label>` — Apple's uids are globally unique, so 5,015 files minted ~30k distinct entity paths. Each distinct path is a dataset schema column, and the OSS server re-merges the cumulative schema per registered file (linear-scan field lookups in arrow-schema) → ~cubic registration cost. The gt layer burned >60 min of CPU without finishing while every other layer took 1.5–3 min.

Fix: shared slot paths `box_00..box_NN` + uid preserved as an `AnyValues` component. Measured: **gt registration >60 min (killed) → 6.5 s** at 5,015 files; identical rendering. Regression test pins the invariant. Standalone repro + upstream issue text: `rerun-io/rerun-schema-width-register-repro`.

### Catalog registration hardening (`ingest/catalog.py`)
- `SKIP` instead of `REPLACE`: REPLACE invalidates the server's schema cache → O(all-sources) recompute per file. Verified in the 0.34.1 sources (`dataset.rs:595-617`, `Overwritten => None`) **and** benchmarked: re-registering 400 real files takes 2.0 ms/file with SKIP vs 23.9 ms/file with REPLACE (12×, growing with dataset size).
- Poll-through-client-drop resume: on big calls the client gRPC connection drops while the server keeps going and commits — the tool polls the segment table instead of resubmitting (duplicate checks run *after* the expensive work), with a stall guard and a 2 h dead-server deadline.
- Layer-major (`<layer>/<video_id>.rrd`) auto-detect alongside the legacy layout; exact `rerun_layer_names` segment-table column (matches prompt-da).
- Hard verification: per-segment layer completeness, not segment counts ("5,015 segments" once hid 4,500 base-only segments).

### Local flow simplified to abc-130k shape (−1,082 lines)
The 612-line local orchestrator (`pipeline.py` + tests + shim + task) is deleted: corpus scale is Modal-only, and the local reference flow is the three existing tools — `download.py --video-ids …`, `ingest_batch.py`, `register_catalog.py` — documented in the README. Validated verbatim: 10 sequences → download 6.6 GB → ingest → register, 10 segments with all 7 layers on a fresh catalog in ~8.5 min.

### Docs
- `docs/full-run-runbook.md`: the 8 steps with their lessons, the three HF rate-limit surfaces, a storage-destination cheat sheet, the gotcha list — hardware-neutral wording (roles, not machines).
- `docs/architecture.md`: schema/lifecycle diagrams updated to the fixed gt layout and SKIP semantics; schema-width bug added to the found-bugs list.
- Parallel-review + Karpathy passes applied: stale docstrings fixed (staging-volume/layer-first reality), dangerous leftover `prefix="rrd"` defaults removed, per-sequence nvenc probe deleted (runbook says to distrust it), serve task gained the `ulimit -n 65536` the 35k-file catalog needs.

## Validation
- `lint` / `typecheck` (0 errors; Modal/boto3 stub false-positives baselined) / `deadcode` / `tests` (33 passed) green in `arkitscenes-download-dev`.
- Register tool exercised against the live 5,015-segment catalog (resumes as all-complete, prints the verification line) and a fresh 10-sequence catalog.
- Data shipped and verified byte-identical across local storage / HF (`pablovela5620/arkitscenes-rrd`) / S3.

## Segment-table preview cards (added after initial review round)
- `register_catalog.py` now also writes and registers a **segment-table blueprint** (`DatasetEntry.register_blueprint(uri, segment_table=True)`), so the dataset review renders each row as a 3D preview card (experimental viewer setting "Table cards and blueprints"). Same register step — no extra workflow.
- The card view excludes both AV1 `VideoStream`s, both depth maps, and the confidence map, keeping mesh/boxes/frusta: with ~15 cards visible, puffin showed dav1d decode saturating ~12 cores (~7 FPS). Exclusion measured **16 → 233 FPS** on the identical headless scene.
- `TIMELINE` moved from `ingest/cli.py` to `ingest/paths.py` (blueprint needs the timeline name; avoids a circular import).


packages/arkitscenes-download/arkitscenes_download/ingest/catalog.py
```diff
@@ -36,16 +55,73 @@ class Config:
...
+            layer_name=layer_name,
+            on_duplicate=OnDuplicateSegmentLayer.SKIP,
+        ).wait()
+    except BeartypeException:
+        raise
+    except Exception as error:  # noqa: BLE001 — dropped call; poll server-side progress instead of resubmitting
+        CONSOLE.print(f"{layer_name}: client call dropped ({str(error)[:70]}) — polling server-side progress")
+        deadline: float = time.perf_counter() + POLL_DEADLINE_S
+        last_count, stalls = -1, 0
+        while True:
```
pablo-rerun: would it make sense to add a tqdm like thing here so its obvious whats happenign? either that or rich?

packages/arkitscenes-download/pyrefly-baseline.json
```diff
@@ -0,0 +1,28 @@
+{
```
pablo-rerun, line 1: why do we have this pyrefly baseline? whats the purpose of this?

packages/arkitscenes-download/arkitscenes_download/ingest/catalog.py
pablo-rerun: Added in c5904d6 — the poll loop now renders a rich Progress bar (per-layer, total = expected files), updating on each 30s poll, and switches its label to "server busy (still grinding)" while the server refuses connections.

packages/arkitscenes-download/pyrefly-baseline.json
pablo-rerun: It suppresses two pyrefly false positives in the Modal job files that we can't fix in code: `boto3` ships no type stubs (imported lazily inside the Modal container function) and pyrefly mis-handles the `modal.App.function` ParamSpec decorator. This is the monorepo's documented convention for compiled/untyped deps (see AGENTS.md — simplecv and live-rerun carry the same kind of baseline, wired via `PYREFLY_EXTRA_ARGS` in the feature's activation env). Without it, `pixi run typecheck` would fail on errors we don't control. Regenerated in c5904d6 to be repo-relative and position-accurate.

packages/arkitscenes-download/arkitscenes_download/ingest/catalog.py
```diff
@@ -36,16 +55,73 @@ class Config:
...
+    The server keeps processing (and eventually commits) after the client's
+    gRPC call drops, so on error we poll the segment table until the layer
+    reaches ``len(paths)`` registered segments or stops making progress.
+    """
+    dataset: DatasetEntry = CatalogClient(config.catalog_url).get_dataset(config.dataset_name)
+    expected: int = len(paths)
+    if registered_layer_count(dataset, layer_name) >= expected:
+        CONSOLE.print(f"{layer_name}: already complete")
+        return
+    started: float = time.perf_counter()
```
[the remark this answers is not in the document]
pablo-rerun: Fixed in c5904d6: `registered_layer_count` now filters the segment table to the segment ids implied by the requested paths, so a `--video-ids` subset of unregistered sequences can no longer be skipped by a high dataset-wide count. Verified against the live 5,015-segment catalog (full + subset invocations).

packages/arkitscenes-download/pyrefly-baseline.json
```diff
@@ -0,0 +1,28 @@
...
+      "concise_description": "Cannot find module `boto3`",
+      "severity": "error"
```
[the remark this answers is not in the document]
pablo-rerun: Regenerated in c5904d6: positions now match the sources (boto3 import at line 58) and the `description` fields are stripped to their concise form, removing the machine-specific absolute paths. Suppression verified (`typecheck` 0 errors).

PR #82, opened 2026-08-06
trtkit: one shared PyTorch→ONNX→TensorRT home; onnxruntime via conda-forge

pablo-rerun:
> [!NOTE]
> Split for review: hotfixes in #83 (merge first), independent cleanups in #84. This PR is the trtkit story only.

One shared **`packages/trtkit`** replaces the five per-package TensorRT runner/builder copies (posekit, sapiens2-pose/sapiens-coco133-pose, mamma, prompt-da, wilor-nano). Net −1,700 lines, behavior preserved. See trtkit's README for the design.

- **Runtime**: `TensorRuntime` protocol (CUDA tensors in/out, keyed by binding name) with torch / ONNX Runtime / TensorRT backends; `run_chunked` owns the reused-buffer contract.
- **Hub**: `ensure_engine` builds + caches machine-local engines keyed on ONNX hash, batch profile, precision (incl. new `allow_tf32`), TRT version, and SM. Engines are never committed; ONNX is the portable artifact.
- **Migrations**: the engine spec is now the batch authority (redundant config cross-checks deleted). No wrapper classes — model packages declare constants + a TypedDict and one boundary function; pipelines hold the bare runtime. `coco133_tensorrt_conversion.py` keeps its bespoke builder on purpose (network I/O dtype surgery).
- **onnxruntime → conda-forge** `*_cuda130` (one dep for linux-64 + aarch64, replaces the jetson-ai-lab pypi hack). Validated on the Spark.

> [!WARNING]
> Recreate existing envs (`rm -rf .pixi/envs/<env> && pixi install --frozen -e <env>`): stale pypi wheel files shadow conda's onnxruntime module and segfault the CUDA EP. Documented in pixi.toml.

Testing: lint/typecheck/deadcode/tests green on all touched packages (posekit 18 incl. three-backend parity + CUDA-graph replay; sapiens2 22, coco133 30, mamma 24, wilor 22, prompt-da 8). Engine caches rebuild once (new cache key). Deep GPU E2E gates (mamma golden, `PROMPTDA_TRT_E2E=1`) worth one post-merge run.


packages/trtkit/src/trtkit/onnx_graph.py
```diff
@@ -0,0 +1,123 @@
...
+Model-agnostic transforms on ONNX interchange files: reading static batch
+sizes and rewriting detector graphs to end before their baked-in NMS. Artifact
+*acquisition* (zoo downloads, checkpoint resolution) stays model-side.
+"""
+
+from pathlib import Path
+from typing import Any
+
+
+def strip_detector_nms(onnx_path: Path) -> Path:
```
pablovela5620: this seems to be specific to a network, rather than something that should be in the loadbearing thing. What i mean is that, shouldn't this be a part of where nms is? not in trtkit?

pablo-rerun: Agreed — it encodes MMDeploy detector export patterns, which is model-family knowledge, not transport. Moved to `posekit.artifacts` (beside the OpenMMLab zip fetching it operates on) in e5148ef; trtkit's `onnx_graph` now keeps only the generic `onnx_static_batch_size` probe. README updated to state the boundary: model-family graph surgery stays model-side.

PR #90, opened 2026-08-11
trtkit: tqdm progress bar for TensorRT engine builds

pablo-rerun:
Stacked on #89. The one-time TensorRT engine build (minutes on first run) sat silent behind a single print line; it now shows a live spinner with an elapsed clock:

```
[trtkit] building TensorRT engine (one-time, may take minutes): depth_model_dot_…engine
⠧ TensorRT build (one-time, can take a few minutes) 0:00:36
```

**Why a spinner and not a progress bar** — measured, not aesthetic: a timestamped `IProgressMonitor` trace of the real o3 build (111 s) shows TensorRT emits callbacks for only ~1% of it — a burst at the start, one 110 s window with zero callbacks, a burst at the end. Its log stream is equally silent during tactic timing (188k VERBOSE messages in the first 5 s, then nothing, plus 4–9 % build-time overhead when subscribed). There is no data to drive a fraction-complete display, so any bar would be synthetic. Earlier iterations on this branch tried stacked tqdm bars (freeze under TRT's phase churn), a step-driven single bar (frozen at 8 % for minutes), and an elapsed-vs-recorded-duration bar (machinery for the rare rebuild case); each was rejected against live tmux evidence.

**What ships:**
- A rich spinner context manager wrapping `build_serialized_network` — no `IProgressMonitor`, no locks, no threads of our own; rich's refresh thread animates it, `transient=True` leaves the terminal clean, and non-tty runs disable it (CI logs unchanged).
- `build_seconds` recorded in each engine manifest as provenance.
- trtkit gains a `rich` dependency (pyproject + pixi feature; one-line lock change — every consuming env already carried it).

**Verified:** live tmux builds at o0 and o3 — spinner and clock animate throughout, clean exit, warm-cache runs print nothing; lint/typecheck/deadcode/tests green; `pixi lock --check` passes.

### DIFF
--- a/packages/trtkit/src/trtkit/trt_builder.py
+++ b/packages/trtkit/src/trtkit/trt_builder.py
@@ -18,6 +18,7 @@ from pathlib import Path
 from typing import Any
 
 import torch
+from tqdm import tqdm
 
 DEFAULT_TRT_CACHE_DIR: Path = Path(os.environ.get("TRTKIT_TRT_CACHE", "~/.cache/trtkit/trt")).expanduser()
 """Machine-local engine cache; override with the ``TRTKIT_TRT_CACHE`` env var."""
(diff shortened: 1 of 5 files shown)


packages/trtkit/src/trtkit/trt_builder.py
```diff
@@ -149,7 +154,76 @@ def build_engine(onnx_path: Path, engine_path: Path, config: TrtBuildConfig, *,
             )
     if has_dynamic_batch:
         builder_config.add_optimization_profile(profile)
-    serialized: Any = builder.build_serialized_network(network, builder_config)
+
+    class BuildProgress(trt.IProgressMonitor):
```
pablovela5620: feels like there should be a much simpler way to do this considering we're only using a spinner

pablo-rerun: Agreed — and it went further than simplifying the class: the monitor is deleted entirely in 3a44080. The trace showed the phase text it maintained ("Timing graph nodes 6/73") stays frozen for 99% of an o3 build, so the callbacks carry nothing worth displaying. The display is now a rich spinner context manager wrapping the build call — no IProgressMonitor, no lock, no shutdown plumbing (−71/+16). Verified live: spinner + elapsed clock animate through the build, clean terminal after, gates green.

PR #98, opened 2026-08-15
prompt-da: stream PromptDA inference from the Rerun dataloader

pablo-rerun:
Runs Prompt Depth Anything over ARKitScenes catalog segments straight from the Rerun PyTorch dataloader — the same shape as `mvs-catalog-depth` — and logs the RGB frame, the ARKit LiDAR prompt and the prediction to the viewer.

```bash
pixi run -e prompt-da-stream --frozen prompt-da-catalog-stream
pixi run -e prompt-da-stream --frozen prompt-da-catalog-stream --data.segments 45662951
```

Nothing is written back to the catalog. The existing `prompt_da_arkitscenes` tool keeps ownership of the registered `promptda` layer, and stays on rerun-sdk 0.35.

## What's here

- **`promptda_stream.py`** — one `RerunIterableDataset` per segment carries the AV1 video (NVDEC-decoded on the GPU), the LiDAR prompt depth and the pose/calibration columns on a single fetch query. The collate stacks whichever rows arrived into one device-resident batch for the TensorRT engine.
- **`simplecv/rerun_dataloader.py`** — `SegmentNvdecDecoder` moves out of `mvs` so both packages share one whole-segment GPU decoder. Deliberately not re-exported from `simplecv/__init__.py`: it needs rerun-sdk's `dataloader` extra, which only the catalog lanes install.
- **`prompt-da-stream{,-dev}` envs** on `catalog-common` + `rerun-prerelease`. The decoder needs the 0.36 dataloader API, and two rerun versions cannot share an env.

## Two decisions worth reviewing

**Orientation is inferred from the frame shape, not read from a capture property.** Ingest bakes orientation into the frames *and* the poses, so a stored frame is already upright — the rotation only buys back the aspect ratio that the engine's fixed landscape input would otherwise squash. This works on both taxonomy generations; the cloud corpus still carries the v1 flat properties that the v2-only parsers reject. Against the old property-driven behaviour over all 5015 segments: identical for 4317 (86%), and the 698 that differ get a 180-degree-rotated input, which for the 684 `turns=2` captures means the network now sees the upright frame instead of an upside-down one.

**The prediction gets its own pinhole.** A `DepthImage` is back-projected with its parent pinhole's intrinsics, and the 1008x756 network output matches neither the native frame nor the 256x192 prompt. The entity is viewer-only, so it stays out of ingest's `paths.py`.

## Verification

Viewer screenshots, not logs — landscape, portrait and 180-degree captures all render aligned depth and a coherent 3D backprojection. Measured on the cloud catalog (RTX 5090, 10 Hz sampling): 262 frames at 28.7 fps landscape, 854 at 25.8 portrait, 1234 at 25.8 for the 180-degree capture. Runs under beartype (`PIXI_DEV_MODE=1`) with every jaxtyping shape holding. `lint`, `typecheck` and `deadcode` clean.

Profiled with py-spy and `torch.profiler`: ~42% GPU inference, ~40% data path, 5% Rerun logging, collate 0.4%. The GPU and data stages never overlap, so a prefetch thread is worth ~17% — left out of this PR deliberately.

## Not included

- No automated test yet.
- No like-for-like benchmark against `prompt-da-arkitscenes`: both read the local catalog, and their v2-only property parsers reject this older cloud dataset.
- When rerun-io/reality#3083 lands, `ColumnDecoder.decode` becomes a batch API and `SegmentNvdecDecoder` needs a signature port.

### DIFF
--- a/packages/prompt-da/src/rerun_prompt_da/promptda_stream.py
+++ b/packages/prompt-da/src/rerun_prompt_da/promptda_stream.py
@@ -52,7 +52,7 @@ class PromptDARow(TypedDict):
     The video frame is device-resident (NVDEC); the rest arrive on the CPU.
     """
 
-    video: UInt8[Tensor, "rgb=3 h w"] | None
+    video: UInt8[Tensor, "3 h w"] | None
     depth: UInt16[Tensor, "1 prompt_h prompt_w"]
     k: Float32[Tensor, "k=9"]
     pose_t: Float32[Tensor, "xyz=3"]
@@ -69,7 +69,7 @@ class PromptDABatch:
     """Sampling-grid timestamp per row, in nanoseconds."""
     quarter_turns: int
     """Counter-clockwise quarter turns applied to reach landscape; undo them when logging."""
-    rgb_bhw3: UInt8[Tensor, "b h w rgb=3"]
+    rgb_bhw3: UInt8[Tensor, "b h w 3"]
     """Device-resident landscape RGB frames."""
     prompt_bhw: Float32[Tensor, "b prompt_h=192 prompt_w=256"]
     """Device-resident landscape LiDAR prompt depth, in metres."""
@@ -144,7 +144,7 @@ class PromptDACollate:
         """Stack the rows that carry every field into one device-resident query."""
         frame_indices: list[int] = []
         timestamps_ns: list[int] = []
-        frames_3hw: list[UInt8[Tensor, "rgb=3 stored_h stored_w"]] = []
+        frames_3hw: list[UInt8[Tensor, "3 stored_h stored_w"]] = []
         prompts_hw: list[UInt16[Tensor, "stored_prompt_h stored_prompt_w"]] = []
         k_matrices_33: list[Float32[ndarray, "3 3"]] = []
         world_T_cam_44: list[Float64[ndarray, "4 4"]] = []
@@ -163,7 +163,7 @@ class PromptDACollate:
                 continue
             if min(video.numel(), depth.numel(), k.numel(), pose_t.numel(), pose_q.numel()) == 0:
                 continue
-            frame_chw: UInt8[Tensor, "rgb=3 stored_h stored_w"] = video
+            frame_chw: UInt8[Tensor, "3 stored_h stored_w"] = video
             depth_1hw: UInt16[Tensor, "1 stored_prompt_h stored_prompt_w"] = depth
             k_9: Float32[Tensor, "k=9"] = k
             pose_t_3: Float32[Tensor, "xyz=3"] = pose_t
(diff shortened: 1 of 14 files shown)


packages/prompt-da/src/rerun_prompt_da/promptda_stream.py
```diff
@@ -0,0 +1,199 @@
...
+        self._index_start_ns: int = segment.index_start
+        self._ns_per_sample: int = ns_per_sample
+        self._device: torch.device = device
+        self._grid_index: int = -1
+
+    def __call__(self, batch: list[PromptDARow]) -> PromptDABatch | None:
+        """Stack the rows that carry every field into one device-resident query."""
+        frame_indices: list[int] = []
+        timestamps_ns: list[int] = []
+        frames_3hw: list[UInt8[Tensor, "rgb=3 stored_h stored_w"]] = []
```
pablo-rerun: the name rgb=3 here doesn't make sense, shouldn't htis be c or channel? why is rgb here?

packages/simplecv/simplecv/rerun_dataloader.py
```diff
@@ -0,0 +1,149 @@
+"""GPU video decoding for the Rerun catalog dataloader.
```
pablo-rerun, line 1: We should add a TODO here because this is a hack while we wait for proper torchcodec support from rerun

packages/prompt-da/src/rerun_prompt_da/promptda_stream.py
pablo-rerun: Agreed, and it was worse than a naming quibble: `rgb=3` appeared only in the new files, while `preprocess_batch` — which this feeds directly — declares `"b h w 3"`. Repo-wide the channel axis is a bare literal (150+ sites vs 7 `c=3` in mvs), so all 7 uses are now plain `3` and the two signatures read identically. Fixed in 2024634.

packages/simplecv/simplecv/rerun_dataloader.py
pablo-rerun: Added a `TODO(rerun#upstream)` in the module docstring. It records that this is a stopgap, what it costs us today (holding a `DatasetEntry` makes it unpicklable, so `num_workers` must stay 0, and the cached CUDA decoder dies with `decode_threads > 1`), and the two upstream PRs that would retire it: reality#2893 (torchcodec decoder, slower than the av path today by its author's measurement) and reality#3083 (batch decode API whose `_DecoderSession` reuses a codec context across a GOP — exactly what this hand-rolls). #3083 also changes the `ColumnDecoder.decode` signature, so the port is flagged. Fixed in 2024634.

PR #99, opened 2026-08-16
shared libs: gauss-surf groundwork across simplecv, trtkit, monoprior, prompt-da

pablo-rerun:
Shared-library groundwork for the gauss-surf Gaussian-splat pipeline (stack 1/5). Everything here is consumed by the later PRs but stands alone.

- **monoprior**: MoGe-v2 TRT surface-normal model, with ONNX export in a subprocess worker (isolates a beartype/SymInt incompatibility at the export boundary; documented in-code). ~9 img/s at 756×1008 on a 5090.
- **trtkit**: `sweep_stale_onnx_exports` — versioned-ONNX stale-file sweep, hoisted from two near-identical caller-local copies (MoGe, PromptDA).
- **prompt-da**: per-segment PromptDA generation + catalog registration (PromptDA layers are generated per segment, ~86 s each); TRT engine touch-ups.
- **simplecv**: `InverseDistortionCoefficients` component type; batch subclasses the forward type overriding only the descriptor, so the wire layout is identical.

**Testing**: trtkit-dev full gates; monoprior + prompt-da focused suites; simplecv descriptor characterization. All green on the assembled branch (see stack PR #103 for the environment that runs everything).

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/packages/monoprior/monopriors/models/surface_normal/_moge_v2_onnx_worker.py
+++ b/packages/monoprior/monopriors/models/surface_normal/_moge_v2_onnx_worker.py
@@ -1,0 +1,30 @@
+"""Uninstrumented subprocess entry point for dynamic-batch MoGe ONNX export."""
+
+import sys
+from pathlib import Path
+from typing import cast
+
+from monopriors.models.surface_normal.moge_v2_trt import Encoder, export_moge_v2_normal_onnx
+
+
+def main() -> None:
+    """Parse the parent export request and build its ONNX artifact."""
+    if len(sys.argv) != 7:
+        raise ValueError(f"Expected six MoGe v2 ONNX worker arguments, got {len(sys.argv) - 1}.")
+    encoder: Encoder = cast(Encoder, sys.argv[1])
+    height: int = int(sys.argv[2])
+    width: int = int(sys.argv[3])
+    resolution_level: int = int(sys.argv[4])
+    max_batch_size: int = int(sys.argv[5])
+    cache_dir: Path = Path(sys.argv[6])
+    export_moge_v2_normal_onnx(
+        encoder=encoder,
+        image_hw=(height, width),
+        resolution_level=resolution_level,
+        max_batch_size=max_batch_size,
+        cache_dir=cache_dir,
+    )
+
+
+if __name__ == "__main__":
+    main()
--- a/packages/monoprior/monopriors/models/surface_normal/_moge_v2_onnx_worker.py
+++ b/packages/monoprior/monopriors/models/surface_normal/_moge_v2_onnx_worker.py
@@ -1,30 +1,41 @@
 """Uninstrumented subprocess entry point for dynamic-batch MoGe ONNX export."""
 
-import sys
+from dataclasses import dataclass
 from pathlib import Path
-from typing import cast
+
+import tyro
 
 from monopriors.models.surface_normal.moge_v2_trt import Encoder, export_moge_v2_normal_onnx
 
 
-def main() -> None:
-    """Parse the parent export request and build its ONNX artifact."""
-    if len(sys.argv) != 7:
-        raise ValueError(f"Expected six MoGe v2 ONNX worker arguments, got {len(sys.argv) - 1}.")
-    encoder: Encoder = cast(Encoder, sys.argv[1])
-    height: int = int(sys.argv[2])
-    width: int = int(sys.argv[3])
-    resolution_level: int = int(sys.argv[4])
-    max_batch_size: int = int(sys.argv[5])
-    cache_dir: Path = Path(sys.argv[6])
+@dataclass(frozen=True, slots=True)
+class Config:
+    """Export request forwarded by the parent predictor process."""
+
+    encoder: Encoder
+    """DINOv2 encoder size."""
+    height: int
+    """Static network height."""
+    width: int
+    """Static network width."""
+    resolution_level: int
+    """Detail level from 0 through 9."""
+    max_batch_size: int
+    """Largest batch encoded in the dynamic ONNX constraint."""
+    cache_dir: Path
+    """Cache root; the graph lands in ``cache_dir / "onnx"``."""
+
+
+def main(config: Config) -> None:
+    """Build the requested ONNX artifact in this clean interpreter."""
     export_moge_v2_normal_onnx(
-        encoder=encoder,
-        image_hw=(height, width),
-        resolution_level=resolution_level,
-        max_batch_size=max_batch_size,
-        cache_dir=cache_dir,
+        encoder=config.encoder,
+        image_hw=(config.height, config.width),
+        resolution_level=config.resolution_level,
+        max_batch_size=config.max_batch_size,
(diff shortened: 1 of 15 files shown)


packages/monoprior/monopriors/models/surface_normal/_moge_v2_onnx_worker.py
```diff
@@ -0,0 +1,30 @@
+"""Uninstrumented subprocess entry point for dynamic-batch MoGe ONNX export."""
+
+import sys
+from pathlib import Path
+from typing import cast
+
+from monopriors.models.surface_normal.moge_v2_trt import Encoder, export_moge_v2_normal_onnx
+
+
+def main() -> None:
```
pablo-rerun: why are you using sys args instead of tyro here like the rest of the mono repo does?

packages/monoprior/monopriors/models/surface_normal/moge_v2_trt.py
```diff
@@ -0,0 +1,377 @@
...
+
+
+class MoGeV2NormalOutput(NamedTuple):
+    """Owning batched normal and mask tensors returned by both predictors."""
+
+    normals_bhw3: Float32[Tensor, "b h w 3"]
+    mask_bhw: Float32[Tensor, "b h w"]
+
+
+class _MoGeV2NormalHeads(torch.nn.Module):
```
pablo-rerun, line 48: is there a reason to do this rather than just fully converting moge? wouuldn't that be more useful? to have the full moge output rather than aribrtraily limiting it to just surface normals? or is there a good reason

packages/monoprior/monopriors/models/surface_normal/moge_v2_trt.py
```diff
@@ -0,0 +1,377 @@
+"""Batched GPU predictors for MoGe v2 camera-space surface normals.
```
pablo-rerun, line 1: why aren't we using trtkit here? it feels like this would be the perfect spot for it?

packages/monoprior/monopriors/models/surface_normal/_moge_v2_onnx_worker.py
pablo-rerun: It's a private subprocess protocol rather than a CLI: `export_moge_v2_normal_onnx` spawns this worker with a fixed 6-arg vector in a `PIXI_DEV_MODE=0` interpreter (to escape beartype's import hook during the dynamo export — the batch dim is a `torch.SymInt`, which correctly fails beartype's eager `int` checks). No human ever invokes it, so tyro's affordances (help text, defaults, completion) go unused. The repo's tyro convention is for `tools/` shims — user-facing CLIs. Happy to switch it to a tyro dataclass anyway if you prefer consistency over the distinction; it's a 5-line change.

packages/monoprior/monopriors/models/surface_normal/moge_v2_trt.py
pablo-rerun: Two deliberate reasons, one honest YAGNI:

1. **The full MoGe pipeline doesn't graph-export.** The network heads (points/depth/normal/mask) export fine, but MoGe's metric-depth story — the affine-invariant point-map postprocess and the `recover_focal_shift` intrinsics solve — is data-dependent logic *outside* the graph. A "full MoGe" engine would still need all of that reimplemented on the output side; it isn't just adding output names.
2. **The consumer doesn't want MoGe depth.** In the gauss-surf pipeline, metric depth comes from PromptDA (prompted by ARKit lidar — strictly better than MoGe's recovered scale for this data); MoGe contributes only normals+mask. Extra heads cost decode compute, engine size, and D2H for outputs nothing reads.
3. YAGNI: extending is cheap when a consumer shows up — same wrapper with more `output_heads`, bump `ONNX_EXPORT_VERSION`. If we want a full-MoGe TRT predictor (points for seeding, say), I'd do it as its own follow-up with its own cache identity rather than pre-paying here.

packages/monoprior/monopriors/models/surface_normal/moge_v2_trt.py
pablo-rerun: It is trtkit end-to-end — the imports are just deferred to the call sites, so they don't show at the top of the diff: `export_onnx` + `sweep_stale_onnx_exports` inside `export_moge_v2_normal_onnx` (L185), and `TrtBuildConfig`/`ensure_engine`/`TensorRtRuntime` inside `MoGeV2TrtNormalPredictor.__init__` (L347-348). Deferred on purpose: importing trtkit pulls the tensorrt runtime, which is heavy, and the torch-only twin predictor (plus anything that imports this module for the tensor contract) shouldn't pay that import cost. Same pattern PromptDA's TRT path uses.

packages/monoprior/monopriors/models/surface_normal/_moge_v2_onnx_worker.py
pablo-rerun: Done in a22b51f — worker is now a tyro `Config` dataclass CLI (still module-private; the parent passes `--encoder/--height/...` flags).

packages/monoprior/monopriors/models/surface_normal/moge_v2_trt.py
pablo-rerun: TODO added in a22b51f on `_MoGeV2NormalHeads`: full-graph export (points+depth heads) when a consumer needs them, with the note that the metric recovery stays outside the graph either way.

packages/monoprior/monopriors/models/surface_normal/moge_v2_trt.py
pablo-rerun: Audited against the repo's canonical TRT flow (prompt-da's `trt_engine.py` + trtkit's shared recipe) — the structure matches on everything that matters (shared `export_onnx` recipe, checkpoint-revision + version cache identity, adapter-wrapper pattern, `ensure_engine` machine-local builds), but your instinct found one genuinely off thing: **the published ONNX graph referenced a permanent pid-suffixed external-data sidecar** (`<name>.onnx.part{pid}.data`). trtkit's atomic publish renamed the protobuf but dynamo's >2 GB weights sidecar kept its temp name forever, and this module had to reverse-engineer that name into `keep_paths`. Fixed at the right layer in a22b51f: `trtkit.export_onnx` now exports into a pid-unique temp *directory* under the final filename and moves both artifacts into place, so the sidecar is born as the deterministic `<name>.data` the graph references (sidecar moved before protobuf so a visible graph never lacks its data; sweep handles abandoned temp dirs; covered by new publish/failure/sweep tests). The other deviation — trtkit imported at call sites here vs module-top in prompt-da — is deliberate (this module doubles as the tensor-contract + torch-twin home, and `import trtkit` pulls the tensorrt runtime); now documented in-code.

PR #121, opened 2026-08-27
dataforge: wildcap dataset — bare self-collected exo/ego mp4s → exoego:v2 base layer

pablo-rerun:
Stacks on #119. Adds `wildcap`, the zero-metadata entry point to the exoego:v2 schema for self-collected exo/ego footage that arrives as bare video files.

## Contract

A capture is just two directories — `<name>/exo/*.mp4` (one file per static camera) plus `<name>/ego/*.mp4` (streams from one head-mounted device). No calibration, no IMU, no per-frame CSVs, no sync manifest. `download` only verifies the tree exists (the corpus is user-assembled); `convert` writes one base rrd per capture.

The base layer asserts only what the mp4s themselves know:

- Videos land on the canonical `rig_NN/cam_MM/pinhole/video` paths, but the `pinhole` node carries **no** `Pinhole` — a calibration layer logs intrinsics at the same path later with zero path churn.
- No root `ViewCoordinates` and no transforms: nothing is posed, so the localization layer that first establishes a world frame owns both.
- The clock is each file's as-encoded PTS, unshifted; mutual sync is a later layer's job. Footage cut to a common start lines up as-is.

Topology mirrors selfcap: exo mp4 N becomes rig N cam 0 (stem order), all ego mp4s share one moving rig as cam 00…. The embedded blueprint (ego row over exo row, 2D only) and the logging loops derive from the same two sorted file lists, so the blueprint can only describe cameras that really get logged. `.mov` files are warned about and skipped (remux first).

`log_video_stream` gains an optional NVENC escape hatch: a B-frame HEVC source forces Mp4Reader into an FFmpeg re-encode, and `DATAFORGE_FFMPEG` + `try_gpu` route that unavoidable re-encode through an NVENC-capable binary. Pass-through sources are untouched.

## Verified

- lint / typecheck / deadcode / 50 tests green.
- End-to-end on self-collected exoego captures (OAK ego trio + four named static exo cameras): 3 episodes converted and registered as 3 segments with a `base` layer on the local catalog, matching selfcap's shape.
- Converted rrds structurally verified: correct static components on rig/cam nodes, video-only `video_time` timeline, no stray Pinhole/ViewCoordinates/transforms, blueprint embedded.

### DIFF
--- a/packages/dataforge/dataforge/apis/register.py
+++ b/packages/dataforge/dataforge/apis/register.py
@@ -41,15 +41,18 @@ def main(config: Config) -> None:
     ).wait()
 
     dataforge_dataset: DataforgeDataset = dataset_config.setup()
-    # A re-register must never truncate an .rbl that a live catalog server still holds open.
+    # Blueprints register once: every register_blueprint call adds a NEW entry to the
+    # dataset's blueprint list (cluttering the viewer's selector), so an incremental
+    # re-register skips them. To refresh a blueprint, delete the dataset and re-register.
+    # A re-register must also never truncate an .rbl a live catalog server holds open.
     blueprint: rrb.Blueprint | None = dataforge_dataset.default_blueprint()
-    if blueprint is not None:
+    if blueprint is not None and dataset.default_blueprint() is None:
         blueprint_path: Path = paths.blueprint_path(paths.output_root(), name)
         with writing.atomic_write(blueprint_path) as temp_path:
             blueprint.save(name, str(temp_path))
         dataset.register_blueprint(blueprint_path.resolve().as_uri(), set_default=True)
     table_blueprint: rrb.Blueprint | None = dataforge_dataset.table_blueprint()
-    if table_blueprint is not None:
+    if table_blueprint is not None and dataset.default_segment_table_blueprint() is None:
         table_path: Path = paths.blueprint_path(paths.output_root(), name, segment_table=True)
         with writing.atomic_write(table_path) as temp_path:
             table_blueprint.save(name, str(temp_path))
--- a/packages/dataforge/dataforge/apis/register.py
+++ b/packages/dataforge/dataforge/apis/register.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 from dataclasses import dataclass, field
 from pathlib import Path
 
-import rerun.blueprint as rrb
 from rerun.catalog import CatalogClient, DatasetEntry, OnDuplicateSegmentLayer
 
 from dataforge import paths, writing
@@ -33,28 +32,27 @@ def main(config: Config) -> None:
         raise FileNotFoundError(f"no {paths.BASE_LAYER}-layer rrds for {name} under {layer_root}")
 
     client: CatalogClient = CatalogClient(config.catalog_url)
-    dataset: DatasetEntry = client.create_dataset(name, exist_ok=True)
-    dataset.register(
+    entry: DatasetEntry = client.create_dataset(name, exist_ok=True)
+    entry.register(
         [path.resolve().as_uri() for path in rrd_paths],
         layer_name=paths.BASE_LAYER,
         on_duplicate=OnDuplicateSegmentLayer.SKIP,
     ).wait()
 
-    dataforge_dataset: DataforgeDataset = dataset_config.setup()
+    dataset: DataforgeDataset = dataset_config.setup()
     # Blueprints register once: every register_blueprint call adds a NEW entry to the
-    # dataset's blueprint list (cluttering the viewer's selector), so an incremental
-    # re-register skips them. To refresh a blueprint, delete the dataset and re-register.
-    # A re-register must also never truncate an .rbl a live catalog server holds open.
-    blueprint: rrb.Blueprint | None = dataforge_dataset.default_blueprint()
-    if blueprint is not None and dataset.default_blueprint() is None:
+    # catalog dataset's blueprint list (cluttering the viewer's selector), so an
+    # incremental re-register skips a blueprint the catalog already has a default for.
+    # To refresh a blueprint, delete the dataset and re-register. A re-register must
+    # also never truncate an .rbl a live catalog server holds open.
+    if entry.default_blueprint() is None:
         blueprint_path: Path = paths.blueprint_path(paths.output_root(), name)
         with writing.atomic_write(blueprint_path) as temp_path:
-            blueprint.save(name, str(temp_path))
-        dataset.register_blueprint(blueprint_path.resolve().as_uri(), set_default=True)
-    table_blueprint: rrb.Blueprint | None = dataforge_dataset.table_blueprint()
-    if table_blueprint is not None and dataset.default_segment_table_blueprint() is None:
+            dataset.default_blueprint().save(name, str(temp_path))
+        entry.register_blueprint(blueprint_path.resolve().as_uri(), set_default=True)
+    if entry.default_segment_table_blueprint() is None:
         table_path: Path = paths.blueprint_path(paths.output_root(), name, segment_table=True)
         with writing.atomic_write(table_path) as temp_path:
-            table_blueprint.save(name, str(temp_path))
-        dataset.register_blueprint(table_path.resolve().as_uri(), segment_table=True)
+            dataset.table_blueprint().save(name, str(temp_path))
+        entry.register_blueprint(table_path.resolve().as_uri(), segment_table=True)
     print(f"registered {len(rrd_paths)} {paths.BASE_LAYER}-layer rrds into '{name}' at {config.catalog_url}")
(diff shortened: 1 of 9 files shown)


pablo-rerun:
Design update since the PR body was written: a wildcap catalog dataset now holds exactly one topology family, because the catalog applies a single default blueprint to every segment of a dataset. The registry entry is `wildcap-selfcollected` (the converter stays generic; another corpus — e.g. a 32-camera mocap stage — gets its own subclass with its own name/root). The dataset default blueprint is derived from the corpus at register time (camera counts + ffprobe'd resolutions; grids of 8/4/2 per tab page for 1080p/2K/4K), and `dataforge-register` now registers blueprints only once instead of accumulating a new copy per run.

packages/dataforge/dataforge/apis/register.py
```diff
@@ -41,15 +41,18 @@ def main(config: Config) -> None:
     ).wait()
 
     dataforge_dataset: DataforgeDataset = dataset_config.setup()
-    # A re-register must never truncate an .rbl that a live catalog server still holds open.
+    # Blueprints register once: every register_blueprint call adds a NEW entry to the
+    # dataset's blueprint list (cluttering the viewer's selector), so an incremental
+    # re-register skips them. To refresh a blueprint, delete the dataset and re-register.
+    # A re-register must also never truncate an .rbl a live catalog server holds open.
     blueprint: rrb.Blueprint | None = dataforge_dataset.default_blueprint()
```
pablo-rerun: this should never be None right? like why would this ever be None if we're reasing an error?

packages/dataforge/dataforge/apis/register.py
```diff
@@ -41,15 +41,18 @@ def main(config: Config) -> None:
...
-    if blueprint is not None:
+    if blueprint is not None and dataset.default_blueprint() is None:
         blueprint_path: Path = paths.blueprint_path(paths.output_root(), name)
         with writing.atomic_write(blueprint_path) as temp_path:
             blueprint.save(name, str(temp_path))
         dataset.register_blueprint(blueprint_path.resolve().as_uri(), set_default=True)
     table_blueprint: rrb.Blueprint | None = dataforge_dataset.table_blueprint()
-    if table_blueprint is not None:
+    if table_blueprint is not None and dataset.default_segment_table_blueprint() is None:
```
pablo-rerun: same here

packages/dataforge/dataforge/datasets/__init__.py
```diff
@@ -7,9 +7,18 @@
 from dataforge.datasets.base import DataforgeDatasetConfig
 from dataforge.datasets.robocap import RobocapConfig
 from dataforge.datasets.selfcap import SelfcapConfig
+from dataforge.datasets.wildcap import WildcapConfig
 
-dataset_defaults: dict[str, DataforgeDatasetConfig] = {config.name: config for config in (RobocapConfig(), SelfcapConfig())}
-"""Every dataset dataforge knows about, keyed by its own ``name`` (the single source of truth)."""
+dataset_defaults: dict[str, DataforgeDatasetConfig] = {
+    RobocapConfig.name: RobocapConfig(),
```
pablo-rerun: this inconsistency seems weird. one is Robocap config.name and the other is a str, we should think about this and make it consistent either by changing wildcap or by canging robocapconfig/selfcapconfig. Tell me the pros/cons of each

packages/dataforge/dataforge/datasets/wildcap.py
```diff
@@ -0,0 +1,263 @@
...
+
+    Higher-resolution streams stay legible in fewer, larger panes: 8 panes for
+    1080p and below, 4 for anything sharper (2 per page felt too sparse). A
+    group larger than its page becomes tabs of grid pages (see
+    ``build_blueprint``).
+    """
+    return 8 if max_height <= 1080 else 4
+
+
+def video_height(video_path: Path) -> int | None:
```
pablo-rerun, line 59: what if the video height is different between the different streams? also using ffprobe the best way to do this? waht about pyav? or torchcodec?

packages/dataforge/dataforge/apis/register.py
pablo-rerun: Right — and the confusion was mine: two different things were both called `default_blueprint`. The left side was the *dataforge hook* (could return None); the right side is the *catalog entry's* registered default (None = nothing registered yet, which is the register-once check). Fixed in 2b2a037: the hooks now return a `Blueprint` (non-optional — wildcap raises `FileNotFoundError` on a corpus with no readable captures, e.g. a wrong `--root`, instead of silently registering a dataset without blueprints), the None guards are gone, and the catalog object is renamed `entry` so `entry.default_blueprint()` vs `dataset.default_blueprint()` read as the two different things they are.

packages/dataforge/dataforge/apis/register.py
pablo-rerun: Same fix (2b2a037): `table_blueprint()` is non-optional now, and the only remaining check is `entry.default_segment_table_blueprint() is None` — the catalog-side "not registered yet" test.

packages/dataforge/dataforge/datasets/__init__.py
pablo-rerun: Agreed it's inconsistent. The root cause: the registry key is really the **CLI subcommand**, and wildcap proved that's a different concept from the **dataset name** (one subcommand, many `wildcap-<corpus>` datasets). Three ways to make it consistent:

**A. Name both concepts explicitly (change the base contract).** Base gets `command: ClassVar[str]` (CLI key); `name` stays the dataset identity — a ClassVar for robocap/selfcap, derived from `corpus` for wildcap. Registry becomes `{config.command: config for ...}` uniformly.
- Pros: both concepts are named, no literals in the registry, the pyrefly `bad-override` suppression on wildcap's `name` property can likely go away (base no longer pins `name` as a ClassVar).
- Cons: fixed datasets carry two attributes that coincide (`command == name`), a touch more ceremony.

**B. Literal keys for everyone (change robocap/selfcap to match wildcap).** `{"robocap": RobocapConfig(), "selfcap": SelfcapConfig(), "wildcap": WildcapConfig()}`.
- Pros: smallest diff; the registry literally reads as the CLI table.
- Cons: key and `config.name` become two sources of truth for fixed datasets and can drift silently — needs a test asserting `key == name` for every non-wildcap entry, and `name`'s docstring ("registry key") becomes a lie.

**C. `name` is an instance field everywhere** (registry keyed by each default instance's name; wildcap derives it in `__post_init__`).
- Pros: one concept only, `--name` works on every verb.
- Cons: robocap/selfcap gain a `--name` flag nobody wants — overriding it makes register's rrd glob find nothing — and wildcap still needs a literal for its `wildcap` subcommand, so the inconsistency just moves.

My recommendation is **A**: it's the only one where the registry stops needing a literal *and* the two concepts stop being conflated, and it removes a type-checker suppression instead of adding a test to guard against drift. Say which and I'll do it.

packages/dataforge/dataforge/datasets/wildcap.py
pablo-rerun: **Mixed heights:** handled one function down — `group_page_size` takes the max over the group, so a mixed group pages for the stream that needs the most pixels. The 32-camera mamma capture is exactly that case (23 streams at 1504 tall, 9 portrait ones at 2048–2056) and lands on 4-up pages. Made the docstring say so explicitly.

**Probe library:** you're right, ffprobe was the wrong tool — shelling out and parsing text is what produced the `2560,` trailing-separator bug. Switched to **PyAV** in 2b2a037: it's already a dependency (declared in `common`, which the dataforge env composes), it reads the header in-process (`container.streams.video[0].height`, ~55 ms vs ~75 ms for the ffprobe subprocess), and failures are typed exceptions instead of stdout parsing. **torchcodec** isn't in the env and would pull torch for a header read — overkill. **cv2** is present, but `VideoCapture` opens a decoder handle to answer a metadata question and its reported dimensions are the less trustworthy of the two for HEVC/rotated phone footage. PyAV is the honest fit.

packages/dataforge/dataforge/datasets/__init__.py
pablo-rerun: Went with **A** in bd3c4c4. Base contract is now `command: ClassVar[str]` (the CLI subcommand / registry key — genuinely class-level) plus a `name` property that defaults to `command`; wildcap overrides `name` as `f"{command}-{corpus}"`. Registry is uniform again — `{config.command: config for ...}`, no literals — and since it's now property-over-property the pyrefly `bad-override` suppression is gone (typecheck reports 0 errors, 0 suppressed). Robocap/selfcap only changed the one ClassVar's name; every `config.name` call site (identity, register glob, catalog name) is unchanged.

PR #129, opened 2026-08-28
zipdepth: live catalog train smoke gate (stack 5/5)

pablo-rerun:
Stack **5/5** — top of the ZipDepth catalog fine-tuning stack. Review bottom-up:

| | PR | what | lines |
|---|---|---|---:|
| 1 | #131 | `arkitscenes-download`: depth-PNG decoders, `landscape_quarter_turns` | +315 |
| 2 | #132 | `zipdepth-catalog` env lane + sample primitives (`targets`, `png`, `threaded`, `stats`) | +900 |
| 3 | #133 | catalog streaming dataset + CUDA builders; throughput probe; holdout eval | +1,027 |
| 4 | #134 | fine-tuning entry point, IO-vs-GPU instrumentation, two vendored trainer lines | +956 |
| 5 | this | live smoke gate; shared `train_smoke` helpers | +269 |

Merge bottom-up and retarget each child to `main` as its parent merges (never `--delete-branch` a stack parent — it closes the child). The whole stack = the former single PR #129, re-cut from its final tree by responsibility (not by commit) after a simplify pass and a structural review; plus, made during the cut: per-module test files, a flat `tools/`, vulture now scanning `zipdepth/catalog`, and one long-lived builder per producer slot instead of `num_producers` new builders (and CUDA streams) per pass.

## This PR
`zipdepth-catalog-train-smoke` is the end-to-end proof for the lane: collect a fixed deterministic probe set from two catalog segments, score the released weights on it, fine-tune 120 steps, assert the probe loss fell, resume, run one `torchrun` leg, and load both checkpoints through `ZipDepthPredictor` before publishing the run directory. `run_required` / `example_rgb` / `publish_latest` move out of the pre-existing `train_smoke` so both smokes share them.

**Review focus**: `collect_probe_batches` (why a windowed mean over the live stream is not a learning signal: later sections of a scene score ~2.5× worse under frozen weights); probe determinism (eval transform, one producer, `shuffle_buffer_size=1`); the `train_smoke.py` refactor belongs to the non-catalog lane.

**Gates**: lint ✓ · pyrefly 0 errors ✓ · deadcode ✓ · 49 passed (catalog env) · 37 passed / 2 skipped (`zipdepth-dev`) ✓
[...]

PR #131, opened 2026-08-28
arkitscenes-download: depth-PNG decoders and landscape_quarter_turns (stack 1/5)

pablo-rerun:
Stack **1/5** for ZipDepth catalog fine-tuning: this → #132 → #133 → #134 → #129. Standalone: nothing in the repo calls these yet; the tests are the contract. Merge bottom-up and retarget each child to `main` as its parent merges (never `--delete-branch` a stack parent — it closes the child).

**What**
- `ingest/depth.py`: the inverse of `encode_depth_png`, kept beside it. `inflate_depth_png_rows` + `decode_depth_png_fast` inflate the encoder's own fixed Up-filter 16-bit grayscale contract with libdeflate (no CRC, no generic filter dispatch) and return `None` for any other PNG; `decode_depth_png` is the general imagecodecs fallback. `encode_depth_png` gains the `UInt16[ndarray, "h w"]` annotation its four callers (gauss-surf, prompt-da ×2, ingest cli) already satisfy.
- `ingest/cells.py`: `landscape_quarter_turns(row)` — the two `capture:orientation*` segment properties → counter-clockwise `rot90` count.
- Tests: PNG round-trip (bit-exact vs imagecodecs), malformed input, orientation, and the existing `table_timestamps` / `match_exact_timestamps` contract from #128 (`table_timestamps` had no in-package consumer, so `deadcode` was red on `main`).

**Review focus**: the chunk walk in `inflate_depth_png_rows` (truncation checks, multi-IDAT concat, IHDR gate, "every filter byte is 2" gate) and the big-endian `.view(">u2")` in `decode_depth_png_fast`.

**Gates** (`arkitscenes-download-dev`): lint ✓ · pyrefly 0 errors ✓ · deadcode ✓ · 74 passed ✓

### DIFF
--- a/packages/arkitscenes-download/arkitscenes_download/ingest/cells.py
+++ b/packages/arkitscenes-download/arkitscenes_download/ingest/cells.py
@@ -19,38 +19,37 @@ def component_instance(value: Any, column_name: str) -> Any:
     return value
 
 
-def landscape_quarter_turns(row: dict[str, Any]) -> int:
-    """Return counter-clockwise quarter turns that bring a stored segment to landscape; 0 for landscape segments.
+def segment_property(row: dict[str, Any], name: str) -> Any:
+    """Return one segment property as a scalar.
 
-    Args:
-        row: Catalog segment row with scalar- or list-shaped orientation properties.
-
-    Returns:
-        Counter-clockwise quarter turns in ``[0, 3]``.
+    Segment rows come from the catalog reader's ``to_pylist()`` as untyped cells whose shape
+    depends on the column: a scalar, a one-element list, a one-element array, or a nested
+    component instance. This is the one place that narrows that shape; callers validate the value.
 
     Raises:
-        KeyError: If a required portrait property is absent.
-        ValueError: If a property is empty or invalid.
+        KeyError: If the property is absent or null.
+        ValueError: If the property cell does not hold exactly one value.
     """
-    orientation_name: str = "property:capture:orientation"
-    if orientation_name not in row or row[orientation_name] is None:
-        raise KeyError(f"segment {row.get('rerun_segment_id', '<unknown>')!r} is missing {orientation_name!r}")
-    orientation: Any = component_instance(row[orientation_name], orientation_name)
-    if isinstance(orientation, (list, np.ndarray)):
-        if len(orientation) != 1:
-            raise ValueError(f"column {orientation_name!r} must contain one value")
-        orientation = orientation[0]
-    if orientation not in ("portrait", "landscape"):
-        raise ValueError(f"column {orientation_name!r} has invalid value {orientation!r}")
-    if orientation == "landscape":
-        return 0
+    if row.get(name) is None:
+        raise KeyError(f"segment {row.get('rerun_segment_id', '<unknown>')!r} is missing {name!r}")
+    value: Any = component_instance(row[name], name)
+    if isinstance(value, (list, np.ndarray)):
+        if len(value) != 1:
+            raise ValueError(f"column {name!r} must contain one value")
+        value = value[0]
+    return value
 
-    turns_name: str = "property:capture:orientation_quarter_turns_ccw"
-    if turns_name not in row or row[turns_name] is None:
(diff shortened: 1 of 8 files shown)


packages/arkitscenes-download/arkitscenes_download/ingest/cells.py
```diff
@@ -17,3 +17,40 @@ def component_instance(value: Any, column_name: str) -> Any:
...
+        row: Catalog segment row with scalar- or list-shaped orientation properties.
+
+    Returns:
+        Counter-clockwise quarter turns in ``[0, 3]``.
+
+    Raises:
+        KeyError: If a required portrait property is absent.
+        ValueError: If a property is empty or invalid.
+    """
+    orientation_name: str = "property:capture:orientation"
```
pablo-rerun: is this funciton the cleanest way to do it? also why are we using Any here? any way to simplify this or do you think this is the best way

packages/arkitscenes-download/arkitscenes_download/ingest/depth.py
```diff
@@ -41,6 +45,122 @@ def encode_depth_png(image_hw: np.ndarray, level: int = 4) -> bytes:
     return b"\x89PNG\r\n\x1a\n" + _png_chunk(b"IHDR", ihdr) + _png_chunk(b"IDAT", idat) + _png_chunk(b"IEND", b"")
 
 
+def inflate_depth_png_rows(blob_u8: UInt8[ndarray, "n"]) -> tuple[UInt8[ndarray, "h row_bytes"], tuple[int, int]] | None:
```
pablo-rerun, line 48: is there a reason to handroll this? its okay if there is but this feels like it should be already included in one of our deps? if not then its fine

packages/arkitscenes-download/arkitscenes_download/ingest/cells.py
pablo-rerun: Fair — it was not the cleanest. It duplicated prompt-da's `portrait_from_segment_row` + `orientation_quarter_turns_from_segment_row`, and its two property reads were copy-paste of each other.

Now (f9f8aa7): one `segment_property(row, name)` reader, `portrait_from_segment_row` on top of it, and `landscape_quarter_turns` is three lines. prompt-da's two helpers are deleted and both its call sites (`prompt_da_chosen`, `prompt_da_arkitscenes`) use these — `grid_batches`/`process_segment` now take the landscape quarter-turns directly instead of `(portrait, raw_turns)` and recombining. Net −28 lines in prompt-da; its three helper tests moved here (81 passed; prompt-da lint/tests green).

On `Any`: rows are `to_pylist()` dicts from the catalog reader, and a cell's shape depends on the column — scalar, one-element list, one-element ndarray, or a nested component instance. `Any` is the honest type at that boundary and `segment_property` is the single place that narrows it (callers validate the value). Typing it `object` would force the same `isinstance` dance at every caller.

packages/arkitscenes-download/arkitscenes_download/ingest/depth.py
pablo-rerun: Partly. The general decode **is** a dependency — `decode_depth_png` just below is `imagecodecs.png_decode`, and it is the fallback for anything outside the contract. The hand-rolled part is only this ~40-line chunk walk (signature → IHDR → IDAT concat → IEND); the inflate itself is libdeflate via `imagecodecs.deflate_decode`.

Why it exists: no dependency (imagecodecs, PIL, cv2, torchcodec, pypng) exposes the *filtered* scanlines — the CUDA builder needs them to run the Up-unfilter on the GPU (`unfilter_up_cuda`). And on the loader's hot path a full CPU decode measured 8.4–9.7 ms/frame (torchcodec / imagecodecs / cv2, 1920×1440 uint16) vs ~4.4 ms for inflate-only. PIL's `PngImagePlugin.ChunkStream` could do the walk but re-verifies CRCs and is semi-private, so it would not be shorter.

While here (f9f8aa7): the CPU Up-unfilter row loop in `decode_depth_png_fast` is now one `np.cumsum(…, dtype=uint8)`.

PR #132, opened 2026-08-28
zipdepth: catalog env lane and sample primitives (stack 2/5)

pablo-rerun:
Stack **2/5** (base #131; next #133). The only PR in the stack that touches `pixi.toml` dependencies or `pixi.lock`.

**What**
- New `zipdepth-catalog` / `zipdepth-catalog-dev` envs: `catalog-common` + `rerun-prerelease` + `cuda` + `zipdepth`, plus torchcodec, imagecodecs, ffmpeg ≥ 8.1 build 902 (older builds dlopen-fail on libjxl), nvidia-ml-py, editable `arkitscenes-download`. Registered in `pyrefly.toml`; `PYREFLY_TARGET` and `[tool.vulture] paths` gain `zipdepth/catalog`. `platforms = ["linux-64"]` follows the `prompt-da` (#98) and `posekit` (#109) lanes. No task blocks yet — each lands beside its script.
- `zipdepth/catalog/` primitives with no catalog dependency:
  - `targets.py` — the numerical contract: inverse depth × `INV_DEPTH_SCALE` 6000 as uint16, `[100, 4000] mm` validity → mask, resize-only `build_train_transform` (drops the vendored random crop-zoom on purpose: at 4:3 → 1024×768 it trains a 1.33× zoom domain that eval never sees), `depth_span_ratio` flat-frame statistic, paired CPU/CUDA `build_training_sample` (tested bit-exact on depth/mask, RGB within resize rounding).
  - `png.py` — `unfilter_up_cuda`: PNG Up-filter reversal as an int32 cumsum on the GPU.
  - `threaded.py` — bounded `ShuffleBuffer`; `iter_threaded` with exception transport, stop event, join timeout.
  - `stats.py` — per-stage counters.

**Review focus**: the constants and tolerances in `targets.py` / `test_catalog_targets.py`; the env composition (this lane drops `common` for `catalog-common`).

**Lock**: `pixi.lock` differs from `main` only in the two new environments (checked env-by-env). `pixi lock --check` is red on untouched `main` as well under pixi 0.78 — it re-keys the v7 platform aliases and rewrites the file — so that re-serialization is a separate workspace PR and is not carried here.

**Gates** (`zipdepth-catalog-dev`, GPU): lint ✓ · pyrefly 0 errors ✓ · deadcode ✓ · 24 passed (CUDA tests run, not skipped) ✓ · `zipdepth-dev` 23 passed / 1 skipped ✓

### DIFF
--- a/packages/zipdepth/zipdepth/catalog/targets.py
+++ b/packages/zipdepth/zipdepth/catalog/targets.py
@@ -1,0 +1,297 @@
+"""Helpers for building ZipDepth training targets (numpy/torch/albumentations only)."""
+
+from collections.abc import Callable
+from dataclasses import dataclass
+from typing import TypeAlias
+
+import albumentations as A
+import cv2
+import numpy as np
+import torch
+from jaxtyping import Bool, Float32, Int64, Shaped, UInt8, UInt16
+from numpy import ndarray
+from torch import Tensor
+from torch.nn import functional as F
+
+from zipdepth.data.transforms import AlbumentationsWrapper
+
+MIN_DEPTH_MM: int = 100
+"""Nearest valid pseudo-depth, in millimetres."""
+MAX_DEPTH_MM: int = 4000
+"""Furthest valid pseudo-depth, in millimetres."""
+INV_DEPTH_SCALE: float = 6000.0
+"""Inverse-depth quantization scale; 10 1/m maps to 60000 without uint16 overflow.
+
+The SSI loss is scale-invariant, so only the resulting floating-point exponent
+range matters after the trainer divides the integer target by 256.
+"""
+
+_DepthTransform: TypeAlias = Callable[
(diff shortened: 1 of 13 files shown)


packages/zipdepth/zipdepth/catalog/targets.py
```diff
@@ -0,0 +1,297 @@
...
+
+
+def build_training_sample_cuda(
+    rgb_chw: UInt8[Tensor, "3 h w"],
+    depth_mm_hw: Shaped[Tensor, "h w"],
+    quarter_turns: int,
+    out_hw: tuple[int, int],
+    generator: torch.Generator,
+    policy: AugmentPolicy,
+) -> dict[str, Tensor]:
```
pablo-rerun, line 216: we should be more precise here, either using a typeddict or a dataclass if it makes sense

pablo-rerun:

PR #133, opened 2026-08-28
zipdepth: stream catalog PromptDA samples; throughput probe and holdout eval (stack 3/5)

pablo-rerun:
Stack **3/5** (base #132; next #134). The PR most likely to hide a real defect — budget review time here.

**What**
- `segments.py` — one catalog connection; the PromptDA segment rows (id + orientation properties).
- `builders.py` — `CpuSampleBuilder` / `CudaSampleBuilder`: one decoded frame + one PNG blob → one training sample. The CUDA path runs entirely on a producer-owned stream (libdeflate inflate → `unfilter_up_cuda` → span filter → resize/flip/color on the GPU); frames never leave CUDA.
- `dataset.py` — `CatalogPromptDepthDataset`: per-segment datafusion query at the real chosen-packet timestamps, zero-copy Arrow views of the blob column (`promptda_blob_views`), an NVDEC decoder over the segment's AV1 packets joined by exact timestamp, N whole-segment producer threads (one long-lived builder each) feeding a bounded shuffle buffer; rank sharding for DDP.
- Read-only consumers: `zipdepth-catalog-throughput` (frames/s + per-stage ms) and `zipdepth-eval-catalog` (scale-shift-aligned AbsRel/δ1 against the pseudo-labels — the released-weights baseline the training PR has to beat).

**Review focus**: `promptda_blob_views` offset arithmetic (every `ValueError` guard is load-bearing); CUDA stream discipline in `CudaSampleBuilder` (`wait_stream`, `record_stream`, one `synchronize` per sample); the decode-error path in `_iter_segment` (warn once, count, continue).

**Gates**: lint ✓ · pyrefly 0 errors ✓ · deadcode ✓ · 36 passed (catalog env) · 24 passed / 2 skipped (`zipdepth-dev`) ✓

**Live**, against the local catalog (`arkitscenes-v2`), 2 producers × 2 segments × 30 batches:

| builder | frames/s | blob_decode | augment | PNG fallbacks | decode skips |
|---|---:|---:|---:|---:|---:|
| CUDA | **26.5** | 7.8 ms | **3.7 ms** | 0 | 0 |
| CPU | 19.4 | 9.3 ms | 16.2 ms | 0 | 0 |

Eval baseline, released weights, holdout segment `42447275` @ net 384: **AbsRel 0.0909 / δ1 0.898** (15 frames, stride 60).

PR #134, opened 2026-08-28
zipdepth: fine-tune on catalog pseudo-labels with IO-vs-GPU instrumentation (stack 4/5)

pablo-rerun:
Stack **4/5** (base #133; next #129).

**What**
- `apis/train_catalog.py` (`zipdepth-train-catalog`): fine-tune the released weights on the catalog stream. The four decisions that made it learn: BatchNorm pinned to eval (`pin_batchnorm_eval` — running-stat drift degrades the released model at *any* lr, even 0), peak lr = config/100 (1e-5; 1e-4 diverges after warmup), resize-only augmentation (from stack 2/5), the `min_depth_span` flat-frame filter. An explicit `total_steps` sizes OneCycle exactly and the loader runs endless reshuffled passes (no epoch model). Resume rebases the schedule; minimal DDP.
- `catalog/instrument.py` — `InstrumentedLoader`: separates data-wait from compute per window and logs `io/*` scalars (data_wait_ms, compute_ms, frames_per_s, gpu_util_pct, per-stage ms) to TensorBoard — the live IO-bound-vs-GPU-bound readout.
- Two sanctioned one-line edits in the vendored `training/trainer.py`, documented in `upstream_cli/__init__.py`: forward `batch["mask"]` to the criterion; an optional `skip_batches` override so resume on an unordered stream stops decode-and-discarding batches. `skip_batches=None` keeps upstream behaviour (tested both ways).

**Review focus**: the two vendored hunks against the docstring; `pin_batchnorm_eval` (rebinds the model root's `train`); `InstrumentedLoader.__iter__` (this *is* the step schedule); `_restore_resume_state` (the only cross-rank broadcast); `main()` is the densest ~180 lines in the stack.

**Result this enables** (overnight run, 70k steps, batch 8, ~8 h): holdout 20 segments / 1,299 frames — fine-tuned @768 **AbsRel 0.0864 / δ1 0.9219** vs released @384 0.0935 / 0.9130 (better on 17/20 segments). Loader after the CUDA path: data_wait 0 ms, GPU 90–98 %, 151 f/s.

**Gates**: lint ✓ · pyrefly 0 errors ✓ · deadcode ✓ · 49 passed (catalog env) · 37 passed / 2 skipped (`zipdepth-dev`) ✓

**Live plumbing**: 50-step fine-tune on 2 segments → `latest.pth` / `final_model.pth`, manifests, and the `[catalog io step 50]` line written; the checkpoint loads and scores through eval (@768 AbsRel 0.1515 ≈ the released model's own @768 score of 0.1486 — 50 steps at 1e-5 do not move the model; learning is proven by the 120-step fixed-probe gate in stack 5/5 and by the overnight result above).

### DIFF
--- a/packages/zipdepth/pyproject.toml
+++ b/packages/zipdepth/pyproject.toml
@@ -37,4 +37,4 @@ paths = ["zipdepth/apis", "zipdepth/catalog", "zipdepth/__init__.py", "tests", "
 exclude = ["*/.pixi/*", "*/__pycache__/*"]
 sort_by_size = true
 min_confidence = 60
-ignore_names = ["rr_config"]
+ignore_names = ["allow_tf32", "benchmark", "forward", "rr_config", "train_epoch"]
--- a/packages/zipdepth/tests/test_catalog_instrument.py
+++ b/packages/zipdepth/tests/test_catalog_instrument.py
@@ -1,0 +1,131 @@
+"""Behavioral tests for catalog loader instrumentation."""
+
+from collections.abc import Iterator
+from pathlib import Path
+from threading import get_ident
+from time import sleep
+from typing import Any
+
+import pytest
+import torch
+from torch.utils.data import DataLoader, IterableDataset
+from torch.utils.tensorboard import SummaryWriter
+
+from zipdepth.catalog.instrument import InstrumentedLoader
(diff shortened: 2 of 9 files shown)


pablo-rerun:

PR #162, opened 2026-08-30
monoprior: FastFoundationStereoPredictor + registry wiring for the stereo tools (add-model PR 2/3)

pablo-rerun:
Stacked on the vendor PR.

- `FastFoundationStereoPredictor(device, checkpoint=None, valid_iters=8, max_disp=416)` implementing `BaseStereoPredictor` → `StereoDepthPrediction`; preprocessing mirrors upstream `run_demo.py` (float RGB, `InputPadder(32)`, fp16 autocast, `test_mode=True`, torch GWC volume).
- Checkpoint: official HF `nvidia/c-fast-foundationstereo` rev `9b446878…` — a **pickled whole module** (no safetensors exists). Loaded with a remapping `Unpickler` (`core.*`/`foundation_stereo_ori.*` → vendored package), deep-copied, strict state-dict reload (722 tensors, 17.66 M params). The checkpoint is NAS-pruned and cannot be rebuilt from `cfg.yaml`; the pickle is the architecture spec. Its `args` lack `normalize`, which `forward()` reads: measured `True` → 0.48 % bad1, `False` → 45 %.
- Registry: `STEREO_PREDICTORS` gains `"FastFoundationStereoPredictor"`; `stereo_depth` demo, `stereo_catalog`, and the Gradio app select by predictor name (small `match`, LAS2 defaults unchanged).
- Tests: fast CPU build-from-config test; slow band: ETH3D playground_1l **EPE 0.241 px / bad1 0.48 %** (LAS2-H 1.12 %) and strict-load of the real checkpoint. Warm timing 5090 fp16: ~83–95 ms/frame at 490×941.

Gates: lint / typecheck / deadcode / tests green (92 passed), slow tests 2 passed.

### DIFF
--- a/packages/monoprior/monopriors/apis/stereo_catalog.py
+++ b/packages/monoprior/monopriors/apis/stereo_catalog.py
@@ -235,7 +235,6 @@ def main(config: StereoCatalogConfig) -> None:
         world_T_rig: Float64[np.ndarray, "4 4"] = world_T_rig_n44[min(int(np.searchsorted(pose_times, np.timedelta64(t_ns, "ns"))), len(pose_times) - 1)]
         left_rect, right_rect = rectify_pair(rect, frame_at(config.left_cam, t_ns), frame_at(config.right_cam, t_ns))
         pred: StereoDepthPrediction = predictor(left_rect, right_rect, K_33=K_rect_32, baseline_m=rect.baseline_m)
-        assert pred.depth_meters is not None
         depth_hw: Float32[np.ndarray, "h w"] = np.where(pred.depth_meters > config.max_depth_m, 0.0, pred.depth_meters).astype(np.float32)
         if config.remove_flying_pixels:
             depth_hw = np.asarray(depth_hw * ~depth_edges_mask(depth_hw, threshold=config.depth_edge_threshold), dtype=np.float32)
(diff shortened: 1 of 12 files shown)


packages/monoprior/monopriors/apis/stereo_catalog.py
```diff
@@ -57,8 +57,10 @@ class StereoCatalogConfig:
     """Frames per second sampled from the video timeline."""
     max_seconds: float | None = 120.0
     """Only the first N seconds of the segment; None processes it all."""
+    predictor_name: STEREO_PREDICTORS = "LiteAnyStereoPredictor"
+    """Which stereo predictor to use."""
     model_size: LAS2ModelSize = "m"
```
pablo-rerun: seems to me that we shouldn't be putting modelsize like this in the stereo_catalog? rather it should be inside the specific one like we do with simplecv exoego? like its specific class?

pablo-rerun: Agreed — reworked in d0c96c05 to the exoego pattern: model-specific options now live in per-predictor config dataclasses next to their predictor (`LiteAnyStereoConfig(model_size, max_disp, checkpoint)`, `FastFoundationStereoConfig(valid_iters, max_disp, checkpoint)`, each with `setup(device)`), combined via `stereo_predictor_defaults` → `tyro.extras.subcommand_type_from_defaults` → `OmitSubcommandPrefixes`, mirroring `simplecv/configs/exoego_dataset_configs.py`. `model_size`/`predictor_name` are gone from `stereo_catalog`/`stereo_depth` (and the `match` blocks with them); the tools take one `predictor:` union field, so the CLI is e.g. `monoprior-stereo-catalog … liteanystereo --model-size h` or `… fast-foundationstereo --valid-iters 16`. `STEREO_PREDICTORS`/`get_stereo_predictor` removed. #163 rebased on top; lint/typecheck/deadcode/tests green on both branches, ETH3D smoke unchanged (EPE 0.241 px / bad1 0.48 %).

PR #164, opened 2026-08-30
prompt-da: cast depth to uint16 and downscale fusion RGB on the GPU (stack 6/8)

pablo-rerun:
Stacked on #129 (zipdepth catalog stack). Throughput-only change to the chosen-frame PromptDA batch tool; stored bytes, layer contract and fused mesh are unchanged.

**What was dumb:** every frame shipped float32 depth (11 MB full-res + 3 MB net-res) and the full 1440×1920 RGB (8.3 MB) over PCIe, then did the scale/cast and the fusion downscale on the host — and fusion only ever used the RGB at network resolution.

**Change:** scale+clamp+cast to uint16 on the device before `.cpu()`; `F.interpolate` the RGB to the fusion size on the device; the writer thread loses its `cv2.resize` (and `cv2` import).

**Measured** (live 5090 batch, 243 vs 12-segment samples, uncontended): 17.3 → 21.6 frames/s (+25%); per-frame D2H 22.4 MB → 6.3 MB; A/B on one segment: identical mesh vertex count, RRD 375 → 374 MB (uint16 cast now truncates a float32 product instead of float64 — occasional ±1 mm).

Gates: ruff, pyrefly (0 errors), vulture, 23 passed / 3 skipped.

### DIFF
--- a/packages/prompt-da/src/rerun_prompt_da/apis/arkitscenes_shared.py
+++ b/packages/prompt-da/src/rerun_prompt_da/apis/arkitscenes_shared.py
@@ -7,6 +7,7 @@ import cv2
 import numpy as np
 import open3d as o3d
 import rerun as rr
+import torch
 from arkitscenes_download.ingest.depth import ArkitDepthConfidence
 from beartype.roar import BeartypeException
 from einops import rearrange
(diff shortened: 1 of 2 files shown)


pablo-rerun:

PR #165, opened 2026-08-30
prompt-da: return TSDF arena pages to the OS at each segment boundary (stack 7/8)

pablo-rerun:
Stacked on the GPU-transfer PR. Fixes the ~0.6 GB/segment host-memory growth that OOM-killed the corpus batch at 142 GB anon-rss after 234 segments (2026-08-30 00:52) — and, because the process lived in `paseo.service`'s cgroup, took the catalog server down with it.

**Root cause (bisected by ablation on real segments):** not Rerun (file sink plateaus at ~550 MB), not the NVDEC decoder, not Arrow. Open3D's `ScalableTSDFVolume` churn on the writer thread frees hundreds of MB per segment that glibc keeps resident in per-thread arenas — used space flat at ~484 MiB while free arena space grew to 3.6 GiB over six segments. `gc.collect()` alone does nothing; `malloc_trim(0)` returns ~1 GB per segment.

**Change:** `gc.collect()` + `malloc_trim(0)` (via `ctypes.CDLL(None)`) at the end of `run_segment()`, where the fuser, executor and recording are out of scope. Linux-only, like NVDEC/TensorRT.

**Verified:** full pipeline smoke +0.2 GB over 3 segments (peak 3.8 GB) vs +0.5–1.1 GB/segment before; in production RSS stayed 4–8 GB across 570 segments (was 36 GB after 50 on the unpatched code).

Gates: ruff, pyrefly (0 errors), vulture, 23 passed / 3 skipped.

### DIFF
--- a/packages/prompt-da/src/rerun_prompt_da/apis/prompt_da_arkitscenes.py
+++ b/packages/prompt-da/src/rerun_prompt_da/apis/prompt_da_arkitscenes.py
@@ -5,6 +5,8 @@ local Rerun catalog, completes depth frame-by-frame, fuses a final TSDF mesh,
 and writes a replaceable ``promptda`` layer back to the dataset.
 """
 
+import ctypes
+import gc
 from collections.abc import Iterable, Iterator
 from concurrent.futures import Future, ThreadPoolExecutor
 from dataclasses import dataclass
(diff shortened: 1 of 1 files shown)


pablo-rerun:

PR #166, opened 2026-08-30
docs: catalog-server fragility and a TensorRT dynamic-batch limitation (stack 8/8)

pablo-rerun:
Docs only.

**AGENTS.md** — the shared rerun catalog on :51235 is in-memory and died four times on 2026-08-29/30: three to another agent session's `pkill -x rerun`, once as collateral of an OOM inside its own `paseo.service` cgroup (`KillMode=control-group`). Records the never-kill-Rerun-by-name rule (kill by pid or port pattern), the `rerun-catalog` symlink as an explicit short-term hack, the cgroup mechanism (`setsid`/`nohup` do not escape it; launch long-lived servers/batches via `tmux new-session`), and the real fixes (persistent store, supervisor).

**trtkit README** — new "Known TensorRT limitations": TRT 10.16–11.2 Myelin fails with `CHECK(is_const())` on any graph whose reshape depends on a dynamic batch symbol (ZipDepth-base's `F.unfold` head; verified identical on 10.16.1, 11.0.0, 11.1.0, 11.2.1 on sm_120; head rewrites only move the failure). Not in NVIDIA's release notes or tracker. Workaround: static-batch export, one engine per batch size. Also: fp16 via `model.half()`, not autocast, for strongly-typed builds. Every other trtkit consumer here builds fine with a dynamic profile.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -109,6 +109,41 @@ layer. `arkitscenes-download-serve` uses `ulimit -n 524288`; restart the server
 to inherit it. This is a capacity workaround—the upstream fix is open-on-demand
 files or an LRU descriptor pool.
 
+### The catalog server is in-memory and dies to `pkill -x rerun`
+
+`rerun server` keeps all registrations in memory: any exit — even a graceful
+SIGTERM — loses every dataset, and the whole corpus must be re-registered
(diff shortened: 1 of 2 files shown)


pablo-rerun:

PR #167, opened 2026-08-30
zipdepth: ZipDepth-PromptDA — prompted metric depth distilled from PromptDA (stack 9/9)

pablo-rerun:
## ZipDepth-PromptDA — prompted metric depth distilled from PromptDA

Builds a 6.14 M-parameter student that takes `(hi-res RGB, 192x256 metric LiDAR prompt) -> metric depth`,
trained purely on PromptDA-large pseudo-labels. It is to PromptDA what ZipDepth is to Depth Anything v2.

Design follows PromptDA's mechanism at ZipDepth's budget: per-image prompt normalisation, multi-scale
injection into all four decoder fusion stages through zero-initialised adapters, and a sigmoid +
denormalise head so metric scale comes from the prompt rather than from RGB.

### Results — held-out metric eval, 20 seed-0 segments, 1,452 frames, no alignment

| | AbsRel | delta1 | MAE |
|---|---:|---:|---:|
| Untrained affine reference | 0.1347 | 0.8094 | 0.222 m |
| **ZipDepth-PromptDA (70k steps)** | **0.0215** | **0.9857** | **0.0338 m** |
| improvement | **6.3x** | +0.176 | **6.6x** |

The aligned diagnostic (0.0208) sits essentially on top of the unaligned metric number (0.0215), which is
the point: the model is genuinely metric, not a relative-depth model rescued by alignment.

Per-segment AbsRel spans 0.015-0.033 across all 20 segments; no outliers, no failures.
Training ran 70,000/70,000 clean, loss 0.081 -> 0.0548.

### Resolution check

Same checkpoint evaluated at two resolutions:

| eval resolution | AbsRel | delta1 | MAE |
|---|---:|---:|---:|
| 768x1024 (training res) | 0.0215 | 0.9857 | 3.38 cm |
| 384x512 (2x LiDAR) | 0.0265 | 0.9838 | 4.00 cm |

768 wins by 23% AbsRel. Note the 384 row carries a train/eval mismatch penalty and is somewhat
pessimistic. Architecturally 768 is the right operating point because `decoder.fuse1` lands at exactly
192x256, the prompt's native size; at 384 the finest decoder stage falls *below* the prompt.

### Visual validation

`prompt-da-polycam-compare` runs teacher and student over the same Polycam capture.
Mean |teacher - student| over 105 frames: **33.3 mm**, consistent with the 33.8 mm holdout MAE on data
from a different capture device. Error is confined to depth discontinuities; surface interiors match.

[...]

PR #172, opened 2026-09-01
monoprior: config-first depth-completion model zoo

pablo-rerun:
Stacked on #167. Posekit-style tyro union (`CompletionConfig` = prompt-da | zipdepth-promptda, each with a torch/onnx/tensorrt `backend` subcommand); ONNX exports move beside their models in monopriors; prompt-da's trt modules become compat shims. Also promotes `log_fused_mesh` + metre→uint16-mm rounding helpers to simplecv (migrating ~8 inlined copies) and extracts the shared Polycam batching preamble. Verified end to end: trio TRT teacher 18.5 ms / TRT student 0.75 ms per frame (24.5x).


pablo-rerun:

PR #173, opened 2026-09-01
prompt-da: polycam hard-frame mining eval tool + trainer profiler

pablo-rerun:
Stacked on the completion-zoo PR. Mines the Polycam capture for the frames where the student deviates most from PromptDA in edge regions (top-10% teacher-gradient pixels) and writes them as a standalone eval set (npz + previews + pyserde-typed metrics.json). Sanity check passed: the #1 hard frame is the plant shot (edge dev 199 mm). Also exposes the trainer's existing torch.profiler through train-catalog flags.


pablo-rerun:

PR #174, opened 2026-09-01
catalog: drop redundant client-side sort and overlap per-segment queries

pablo-rerun:
The training pipeline was pipeline-bound, not GPU-bound (py-spy: main thread 36% in next(dataloader); producers 40% in datafusion to_arrow_table). Two fixes, no SDK change needed: (1) remove `.sort(TIMELINE)` — the reader already returns (segment, index)-ordered rows, and the client-side SortExec re-materialized 250–400 MB of blob columns per segment (~4x query wall); a monotonicity guard now enforces the ordering contract loudly. (2) overlap the depth query (worker thread) with the video query + NVDEC open (producer thread — decoder creation needs that thread's CUDA context). Measured: loader 60.8 → 152.6 f/s, segment_query 22.4 → 3.5 ms/frame, sustained training 12 → 17.9 it/s (500k steps: ~11.6 h → ~7.8 h).

### DIFF
--- a/packages/simplecv/simplecv/rerun_dataloader.py
+++ b/packages/simplecv/simplecv/rerun_dataloader.py
@@ -61,13 +61,18 @@ def open_segment_decoder(
         and the decoder over the whole segment.
     """
     view: DatasetView = dataset.filter_segments(segment_id).filter_contents(entity)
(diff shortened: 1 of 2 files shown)


pablo-rerun:

PR #175, opened 2026-09-01
workspace: migrate to rerun-sdk 0.37.0rc2 (gradio packages on a 0.36.3 lane)

pablo-rerun:
Whole-workspace bump to 0.37.0rc2. The `datafusion` extra is removed upstream — pins move to `[catalog]` (+`[dataloader]` on catalog lanes). `rerun-sdk` leaves `common`: everything rides the new `rerun-037` feature except the 13 Gradio env-families, which ride a temporary `gradio-rerun-036` lane (gradio-rerun hard-pins rerun-sdk==0.36.3; collapse recipe + pin-bump holds documented in pixi.toml). Ports: SegmentNvdecDecoder to the 0.37 batch-decode contract (+regression test), fetch_block_size, NoShuffle(), datatypes→encodings, Mp4Reader sparse-keyframe timestamp fix. Rust: re_* crates to 0.37.0-rc.2 (no egui bump needed), builds + tests green. Validated on the upgraded catalog server + re-registered corpus: loader 153 f/s, training 17.2 it/s sustained, trio 0.73 ms TRT / 9.3 ms torch — all pre-migration numbers held.

PR #177, opened 2026-09-02
zipdepth: reallocatable metric loss, edge-stratified eval, simplify sweep with three fixes (stack 14/14)

pablo-rerun:
Stacked on #175 (rerun-037-migration). Two commits from the edge-quality investigation:

- **Reallocatable metric loss + edge-stratified eval** (ee99c99d): per-scale gradient weights, a widenable gradient validity window (builders keep real teacher values beyond the 4 m L1 window; 0 mm stays the invalid encoding), relaxable pooled-mask threshold, optional teacher-edge-weighted L1. Defaults verified bit-equal to the previous objective. `eval-catalog` gains edge/flat stratified MAE against a hole-aware bilinear-prompt baseline.
- **Four-lens simplify sweep** (216d7cd7): MetricDepthLoss/trainer/eval cleanup (bit-equal at defaults, pinned by tests), batch decode grouping per segment in the simplecv decoder, and three real fixes: the loss was fed the windowed `target_valid` so `grad_max_depth_m` could never widen the window (the overnight sweep's wide arms silently trained the default window); edge weights read holes as edges; the DDP smoke leg trained the prompted model.

Context: the overnight loss sweep run with these levers was a negative result (v4 stays production); write-up in the consolidated report https://pablos-4800gt.ilish-ruler.ts.net:8768/zipdepth/zipdepth-promptda-project-report.html. Gates: zipdepth 58 pass, simplecv 151 pass, lint/typecheck/deadcode clean, catalog train smoke green.


pablo-rerun:

PR #178, opened 2026-09-03
workspace: exo-calib envs, linux-aarch64 lanes, onnxruntime-gpu 1.29 from PyPI (exo-calib stack 1/6)

pablo-rerun:

This branch contains #185's commit, so its diff shows those 13 files until #185 merges (GitHub will not retarget the bottom PR of a stack).

## What
- `exo-calib` / `exo-calib-dev` envs and the package scaffold (pyproject, `.envrc`, beartype, pyrefly and `.gitignore` registrations). Deps: kornia-rs, monopriors, posekit, trtkit, sam3, vggt; siblings pinned with `[tool.uv.sources]`.
- `linux-aarch64` lanes for simplecv, monoprior, mv-api, mv-api-catalog, sapiens2-pose and posekit, so the DGX Spark runs the whole pipeline.
- onnxruntime moves to the PyPI `onnxruntime-gpu==1.29.0` wheel: the conda build has no sm_121 kernels for the GB10. cuDNN pinned to the 9.x ABI; trtkit preloads the conda-prefix CUDA libraries.
- AGENTS.md: the aarch64 lock-generation limit and the beartype PEP 526 gotcha (exo-calib passes `claw_is_pep526=False`).
- `pixi.lock` regenerated on linux-64.

## Verified
`pixi lock --check` and a no-op `pixi lock` on all three machines; `tests` for exo-calib, mv-api-catalog, simplecv, trtkit, posekit, monoprior and sapiens on the dl-server and the Spark. ORT 1.29 CUDA EP on the GB10: RTMW-x 7 ms, YOLOX 9 ms per run.


AGENTS.md
```diff
@@ -31,7 +31,7 @@ The workspace `platforms` list defines the full platform vocabulary: the plain `
...
...requirements, so it never needed build isolation. A host can regenerate the lock only for envs that have a lane it can run: pixi's build dispatch must instantiate every env containing editable/git PyPI deps on the current machine, and picks any of the env's platforms to do so. Every env has a linux-64 lane, so **a linux-64 host (pablo-dl-server) regenerates the whole-workspace lock**. A linux-aarch64 host (the Spark) regenerates changes scoped to envs with an aarch64 lane (e.g. `exo-calib`, `mv-api`, `simplecv`), but a change to a shared feature (`cuda`, `common`, `dev`) invalidates the PyPI solve of the 26 linux-64-only envs (`gauss-surf`, `dpvo`, `sam3`, …) and `pixi lock` fails with `build dispatch initialization failed: the environment '<env>' does not support 'linux-aarch64' on this machine` (pixi 0.77.1, seen 2026-09-02). Re-lock such changes on the dl-server: `git bundle create` the branch plus `git diff` for uncommitted work, apply in a temporary worktree there, run `CONDA_OVERRIDE_CUDA=13.0 pixi lock`, copy `pixi.lock` back, and confirm with `pixi lock --check` on the Spark. Full lock *regeneration* from macOS fails for the same reason: linux-only envs cannot exist on osx.
```
pablo-rerun: thsi feels overly verbose and historical, It seems like we should make it forward facing rather than saying this specific package got vendered or some other package by name is prebuilt

pixi.toml
```diff
@@ -155,6 +148,22 @@ onnxruntime = { version = ">=1.26.0,<2", build = "*cuda130*" }
 [feature.cuda.pypi-dependencies]
 tensorrt-cu13 = { version = "==11.2.1.2", index = "https://pypi.nvidia.com" }
 nvidia-cuda-runtime-cu13 = "==0.0.0a0"
+# onnxruntime — the PyPI wheel, like TensorRT: conda-forge's onnxruntime build has
```
pablo-rerun: this is way too long a comment

packages/exo-calib/pyproject.toml
```diff
@@ -0,0 +1,42 @@
+[project]
+name = "exo-calib"
+version = "0.1.0"
+authors = [{name = "Pablo Vela", email = "pablovela5620@gmail.com"}]
+requires-python = ">=3.12"
+dependencies = [
+    # catalog-common: numpy, scipy, jaxtyping, rerun-sdk, huggingface_hub,
+    #                 hf-transfer, tqdm, pyserde, av, opencv, simplecv
+    # conda (pixi feature): open3d, torchcodec; pypi: kornia-rs (bundle adjustment)
+    "monopriors",
```
pablo-rerun, line 11: its confusing to have monopriors here and also in the pixi.toml feature def, in particular because I think that theres a pypi monopriors package that could mess things up. Can you triple check this?

AGENTS.md
pablo-rerun: Rewritten as the rule rather than the history: resolution never needs package Python, build dispatch decides which host can lock what, and how to re-lock on a linux-64 host. The package roll-call is gone; the `Former no-build-isolation deps` gotcha further down got the same treatment.

pixi.toml
pablo-rerun: Cut to five lines (why PyPI, why this exact pin, where the CUDA libs come from).

packages/exo-calib/pyproject.toml
pablo-rerun: Checked three ways. The lock records `./packages/monoprior` for both exo-calib envs; the installed distribution in `exo-calib-dev` is the editable path (`direct_url.json` → file:///…/packages/monoprior, v0.3.3, same for posekit / simplecv / sam3); and this is the repo's existing convention for in-repo deps (mv-api lists `monopriors`, posekit lists `simplecv`, monoprior lists `sam3`). You're right that an unrelated `monopriors` 0.1.0 exists on PyPI: the editable path entry in `[feature.exo-calib]` is what pins the name, so it can only bite someone installing the package outside pixi. I added a comment saying exactly that. If you'd rather remove the ambiguity entirely, the repo-wide move is to drop workspace names from `[project].dependencies` everywhere; I kept the convention here.

pablo-rerun:

PR #179, opened 2026-09-03
simplecv: catalog rig-layout parser, typed umeyama result, codec-aware segment decoder (exo-calib stack 2/6)

pablo-rerun:

## What
- `simplecv/catalog_rig_layout.py`: the read-side parser of the exoego:v2 rig layout from a catalog schema (`CatalogRigLayout`, `parse_rig_layout`, `read_rig_kinds`; `RigKind` is exo / ego / quest). Consumers keep only their selection policy.
- `simplecv/catalog_calibration.py`: `read_camera_calibration` (pinhole + transform, column-major transpose), shared by exo-calib and mv-api.
- `simplecv/catalog_video_codec.py`: `catalog_codec_name` maps the `VideoStream:codec` FourCC; `open_segment_decoder` muxes AV1 / H.264 / H.265 from it.
- `umeyama_alignment` returns a `SimilarityTransform` with `apply()` / `as_matrix()` and a ≥3-correspondence guard; the metric exoego calibration drops its private copy.

## Verified
ruff, pyrefly; simplecv tests 171 passed / 29 skipped on the dl-server and the Spark.

### DIFF
--- a/packages/simplecv/simplecv/apis/metric_exoego_calib.py
+++ b/packages/simplecv/simplecv/apis/metric_exoego_calib.py
@@ -33,6 +33,7 @@ from simplecv.data.skeleton.coco_133 import (
     RIGHT_HAND_IDX,
 )
 from simplecv.ops.triangulate import batch_triangulate
+from simplecv.ops.umeyama import umeyama_alignment
 from simplecv.rerun_custom_types import PinholeWithDistortion, Points2DWithConfidence, Points3DWithConfidence, confidence_scores_to_rgb
 from simplecv.rerun_log_utils import RerunTyroConfig, log_pinhole, log_video
 from simplecv.rig import rebuild_camera_with_extrinsics
@@ -237,7 +238,7 @@ def single_frame_mv_hands(
 
     uvc_coco_list: list[Float32[ndarray, "133 3"]] = []
     uvc_cam_names: list[str] = []
-    for ego_cam_name, rgb_hw3 in tqdm(
+    for ego_cam_name, frame_rgb in tqdm(
         list(zip(ego_sequence.ego_cam_dict.keys(), rgb_list, strict=True)),
         desc="Hand detect/keypoints (calib frame)",
         leave=False,
@@ -245,15 +246,15 @@ def single_frame_mv_hands(
         # don't run on the quest cameras
         if "quest" in ego_cam_name.lower() or "rgb" in ego_cam_name.lower():
             continue
-        det_results: DetectionResult = hand_detector(rgb_hw3, hand_conf=0.3)
+        det_results: DetectionResult = hand_detector(frame_rgb, hand_conf=0.3)
         cam_log_path: Path = parent_log_path / "ego" / ego_cam_name
         pinhole_log_path: Path = cam_log_path / "pinhole"
-        rr.log(f"{pinhole_log_path}/image", rr.Image(rgb_hw3).compress(jpeg_quality=90))
+        rr.log(f"{pinhole_log_path}/image", rr.Image(frame_rgb).compress(jpeg_quality=90))
         uvc_coco: Float32[ndarray, "133 3"] = np.full((133, 3), np.nan, dtype=np.float32)
         # log detected hands and keypoints
         if det_results.left_xyxy is not None:
             wilor_preds: FinalWilorPred | KeypointResults = hand_keypoint_model(
-                rgb_hw3=rgb_hw3, xyxy=det_results.left_xyxy, handedness="left"
+                rgb_hw3=frame_rgb, xyxy=det_results.left_xyxy, handedness="left"
             )
             assert isinstance(wilor_preds, KeypointResults)
             if debug:
@@ -272,7 +273,7 @@ def single_frame_mv_hands(
 
         if det_results.right_xyxy is not None:
             wilor_preds: FinalWilorPred | KeypointResults = hand_keypoint_model(
-                rgb_hw3=rgb_hw3, xyxy=det_results.right_xyxy, handedness="right"
+                rgb_hw3=frame_rgb, xyxy=det_results.right_xyxy, handedness="right"
             )
             assert isinstance(wilor_preds, KeypointResults)
             if debug:
@@ -377,12 +378,8 @@ def single_frame_mv_hands(
     # dst_points are GT joints expressed in the Quest/world frame coming from the labels.
     dst_points: Float[ndarray, "n_valid 3"] = gt_xyzc[valid_mask, :3].astype(np.float64, copy=False)
 
-    # Similarity that maps Oak→new-Oak-world (destination ← source).
-    new_oak_world_T_oak_world: Float32[ndarray, "4 4"] = umeyama_transform(
-        src_points=src_points,
-        dst_points=dst_points,
-        allow_scaling=False,
-    )
+    # Rigid transform that maps Oak-world into new-Oak-world, as a homogeneous matrix.
+    new_oak_world_T_oak_world: Float32[ndarray, "4 4"] = umeyama_alignment(src_points, dst_points, allow_scaling=False).as_matrix().astype(np.float32)
     src_points_h: Float[ndarray, "n_valid 4"] = np.concatenate(
         (src_points, np.ones((num_valid, 1), dtype=np.float64)),
         axis=1,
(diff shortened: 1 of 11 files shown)


packages/simplecv/simplecv/apis/metric_exoego_calib.py
```diff
@@ -377,12 +378,12 @@ def single_frame_mv_hands(
...
 
-    # Similarity that maps Oak→new-Oak-world (destination ← source).
-    new_oak_world_T_oak_world: Float32[ndarray, "4 4"] = umeyama_transform(
-        src_points=src_points,
-        dst_points=dst_points,
-        allow_scaling=False,
-    )
+    # Similarity that maps Oak→new-Oak-world (destination ← source), as a homogeneous ``dst_T_src``.
+    similarity: SimilarityTransform = umeyama_alignment(src_points, dst_points, allow_scaling=False)
+    new_oak_world_T_oak_world_f64: Float[ndarray, "4 4"] = np.eye(4, dtype=np.float64)
```
pablo-rerun: I'm noticing that you're putting the name of things such as f64, the point of using jaxtyping is to make it that way. Rather than doing things via the name, use jaxtyping to specify it.

So Float if it can be of any Float32 if it needs to be of 32bit and Float64 if thats wahts needed there. Rather than polluting the name

packages/simplecv/simplecv/ops/umeyama.py
```diff
@@ -0,0 +1,65 @@
...
+        The similarity such that ``dst_points ≈ scale * (rotation_33 @ src_points.T).T + translation_3``.
+
+    Raises:
+        ValueError: If the shapes differ or the source variance is at most ``eps``.
+    """
+    if src_points.shape != dst_points.shape:
+        raise ValueError(f"Source and destination shapes must match; got {src_points.shape} vs {dst_points.shape}")
+    src_f64: Float64[ndarray, "n 3"] = src_points.astype(np.float64, copy=False)
+    dst_f64: Float64[ndarray, "n 3"] = dst_points.astype(np.float64, copy=False)
+    src_mean_3: Float64[ndarray, "3"] = np.mean(src_f64, axis=0)
```
pablo-rerun: again, I dont think you need to put src_mean_3, src_mean is probalby enough considering you already have the shape with jaxtyping. Its also a bit harder to usnderand why you moved things to float64 and did a copy? Is that necessary?

packages/simplecv/simplecv/ops/umeyama.py
```diff
@@ -0,0 +1,65 @@
...
+    reflection_3: Float64[ndarray, "3"] = np.ones(3, dtype=np.float64)
+    if np.linalg.det(u_33) * np.linalg.det(vt_33) < 0.0:
+        reflection_3[2] = -1.0
+    dst_R_src: Float64[ndarray, "3 3"] = u_33 @ np.diag(reflection_3) @ vt_33
+    src_var: float = float(np.mean(np.sum(src_centered**2, axis=1)))
+    if src_var <= eps:
+        raise ValueError(f"Source variance too small for stable Umeyama estimation (variance={src_var})")
+    scale: float = float(np.sum(singular_vals_3 * reflection_3) / src_var) if allow_scaling else 1.0
+    dst_t_src: Float64[ndarray, "3"] = dst_mean_3 - scale * (dst_R_src @ src_mean_3)
+    return SimilarityTransform(rotation_33=dst_R_src, translation_3=dst_t_src, scale=scale)
```
pablo-rerun: would also be nice to have consistent names here like you did with dst_R_src instead of just rotation. That way its clear which way its going

packages/simplecv/simplecv/catalog_rig_layout.py
```diff
@@ -0,0 +1,188 @@
...
+_RIG_NODE_PATH: re.Pattern[str] = re.compile(r"^/world/(rig_\d+)$")
+
+
+@dataclass(frozen=True, slots=True)
+class CatalogComponent:
+    """One component column of a catalog schema."""
+
+    entity_path: str
+    """Entity path with a leading slash, e.g. ``/world/rig_00/cam_00``."""
+    component: str
```
pablo-rerun: is there a way we can make this type stronger? rather than just a str? I'm not sure if rerun exposes this or not but it would be nice if we could make this more precise

packages/simplecv/simplecv/catalog_rig_layout.py
```diff
@@ -0,0 +1,188 @@
...
+    """Component name as ``Archetype:field``, e.g. ``Transform3D:translation``."""
+    is_static: bool
+    """Whether the column is static (logged without a timeline)."""
+
+
+def catalog_components(schema: Any) -> list[CatalogComponent]:
```
pablo-rerun: same here, we really should avoid Any/Object/ect if we can, its hard to tell what it is if we do this unless this is the only way

packages/simplecv/simplecv/catalog_rig_layout.py
```diff
@@ -0,0 +1,188 @@
...
+
+    Args:
+        schema: A Rerun catalog schema exposing ``component_columns()`` (a list or a
+            dict keyed by column descriptors with ``entity_path``, ``component`` or
+            ``name`` / ``archetype`` + ``component_type``, and ``is_static``).
+
+    Returns:
+        One record per column whose entity path and component name could be read.
+    """
+    columns: Any = schema.component_columns()
```
pablo-rerun: lots of anys, I dont know if this is just because of the way rerun does things and if theres no way to have stronger types here

packages/simplecv/simplecv/rerun_dataloader.py
```diff
@@ -41,11 +42,32 @@
 FrameRgbChw: TypeAlias = UInt8[Tensor, "3 h w"]
 """One decoded frame, channels-first RGB on the decoder's device."""
 
+PYAV_CODEC_NAME: dict[rr.VideoCodec, str] = {rr.VideoCodec.AV1: "av1", rr.VideoCodec.H264: "h264", rr.VideoCodec.H265: "hevc"}
+"""PyAV codec name for each catalog ``VideoStream:codec`` value that ``wrap_mp4`` can mux."""
+
+
+def pyav_codec_name(codec_value: int) -> str:
```
pablo-rerun: should this be a string? or a literal for the output type?

packages/simplecv/simplecv/apis/metric_exoego_calib.py
pablo-rerun: Dropped the dtype from the name. The matrix is now built float32 directly (`new_oak_world_T_oak_world: Float32[ndarray, "4 4"]`) from a `SimilarityTransform`-typed estimate; the values are identical to before, since casting the two blocks after the float64 estimate is what the whole-matrix cast did.

packages/simplecv/simplecv/ops/umeyama.py
pablo-rerun: Renamed to `src`, `dst`, `src_mean`, `dst_mean`, `covariance`, `reflection`, `src_variance`; dtype and shape now live only in the jaxtyping annotations. The float64 cast stays and the docstring now says why: the SVD and the scale ratio lose precision in float32 and callers pass float32 detections. `copy=False` makes it a no-op for float64 inputs. Added a test that float32 inputs produce a float64 estimate.

packages/simplecv/simplecv/ops/umeyama.py
pablo-rerun: Fields are now `dst_R_src`, `dst_t_src`, `scale`, and the callers use direction names as well (`gt_T_pred` in exo-calib's eval and report).

packages/simplecv/simplecv/catalog_rig_layout.py
pablo-rerun: Rerun only exposes the identifier as a string (`ComponentColumnDescriptor.component`, e.g. `Transform3D:translation`; `archetype` and `component_type` are strings too, no enum). So I encoded its grammar instead: `CatalogComponent` now carries `archetype: str | None` and `field: str` parsed from it, and the parser matches `archetype == "VideoStream" and field == "sample"` rather than string prefixes. `rig_kind` is `RigKind = Literal["exo", "ego"]`, and exo-calib's reader rejects anything else.

packages/simplecv/simplecv/catalog_rig_layout.py
pablo-rerun: Replaced with two small runtime-checkable Protocols: `SchemaLike` (has `component_columns()`) and `ComponentColumnLike` (`entity_path`, `component`, `is_static` as read-only properties). `rerun.catalog.Schema` and `rerun_bindings.ComponentColumnDescriptor` satisfy them (pyrefly checks the assignment) and the test stand-ins still pass beartype.

packages/simplecv/simplecv/catalog_rig_layout.py
pablo-rerun: Gone. Real descriptors always carry `component`, so the `name` / `archetype`+`component_type` fallbacks (inherited from mv-api's older code) are removed along with the `Any`s; the descriptor is typed through the `ComponentColumnLike` Protocol.

packages/simplecv/simplecv/rerun_dataloader.py
pablo-rerun: `PyAvCodecName = Literal["av1", "h264", "hevc"]`; `pyav_codec_name` returns it and `wrap_mp4` takes it.

pablo-rerun:

pablo-rerun:
Landed on main as #207 (same content, rebased onto main after GitHub's stack tooling closed this PR when its base branch was deleted following the #178 merge).

PR #182, opened 2026-09-03
exo-calib: catalog IO, layer contract, Stage A init, Stage B tracked keypoints (stack 5/6)

pablo-rerun:
**exo-calib stack:** [1/6 workspace](https://github.com/rerun-io/examples-monorepo/pull/178) · [2/6 simplecv](https://github.com/rerun-io/examples-monorepo/pull/179) · [3/6 mv-api](https://github.com/rerun-io/examples-monorepo/pull/180) · [4/6 geometry core](https://github.com/rerun-io/examples-monorepo/pull/181) · [5/6 catalog + Stage A/B](https://github.com/rerun-io/examples-monorepo/pull/182) · [6/6 refine + pipeline](https://github.com/rerun-io/examples-monorepo/pull/183). Merge in order, #185 first.

## What
- `catalog_io`: `StageConfig` (the six flags every stage shares) and `stage_context()` (connect, pick the segment, discover the rig). Exo/ego selection via `simplecv.catalog_rig_layout`; calibration the pipeline wrote never counts as ground truth.
- `layer_io`: the exocalib layer contract (layer names, entity paths, `CalibrationVariant`, camera frusta, `RefinementDiagnostics`, 3D tracks, base-node calibration write). `kp2d_layer`: the Stage B record round-trip.
- `video_io`: one NVDEC decoder per exo camera over catalog packets.
- Stage A `calibrate_init`: one synchronized frame per view → G3T → MoGe-2 metric scale → gravity-checked RANSAC floor → `exocalib_init`.
- Stage B `keypoints2d`: all views in lockstep; boxes projected from the 3D-tracked skeleton with a YOLOX re-anchor every 2 s; RTMW-x (TensorRT) COCO-133 keypoints → `exocalib_kp2d` plus a gated overlay.
- pixi tasks `exocalib-init`, `exocalib-kp2d`.

## Verified
ruff, pyrefly, vulture; 36 tests with beartype on (whole stack). End-to-end numbers are in #183.

PR #181, opened 2026-09-03
exo-calib: geometry core — correspondences, triangulation, BA, focal, confidence, eval (stack 4/6)

pablo-rerun:
**exo-calib stack:** [1/6 workspace](https://github.com/rerun-io/examples-monorepo/pull/178) · [2/6 simplecv](https://github.com/rerun-io/examples-monorepo/pull/179) · [3/6 mv-api](https://github.com/rerun-io/examples-monorepo/pull/180) · [4/6 geometry core](https://github.com/rerun-io/examples-monorepo/pull/181) · [5/6 catalog + Stage A/B](https://github.com/rerun-io/examples-monorepo/pull/182) · [6/6 refine + pipeline](https://github.com/rerun-io/examples-monorepo/pull/183). Merge in order, #185 first.

The numerical core of exo-calib, no IO. Tuned values are named constants in the stage modules (#182, #183); these functions take them as required arguments.

## What
- `correspondences`: Kineo pair rule over all view pairs; under-supported-view detection.
- `triangulation`: batched confidence-weighted DLT, robust pruning with leave-one-out consensus, single-frame multi-view triangulation for the tracker.
- `ba`: kornia-rs Schur bundle adjustment with soft camera-centre priors and re-triangulation rounds; rotation-update guard.
- `focal`: one log-focal scale per camera with poses and points fixed.
- `confidence`, `boxes`, `keypoints`: AssemblyHands-X gating, temporal median filter, reprojection-driven person boxes, the Stage B record.
- `cameras`, `eval`: the rig container, SE(3) / Sim(3) errors through simplecv's umeyama, rig geometry.

## Verified
ruff, pyrefly; tests on one synthetic ring rig (`conftest.py`) with beartype on. Read in order: correspondences → triangulation → ba → focal.

PR #180, opened 2026-09-03
mv-api: catalog prediction tool accepts rig-layout datasets and full skeletons (exo-calib stack 3/6)

pablo-rerun:
**exo-calib stack:** [1/6 workspace](https://github.com/rerun-io/examples-monorepo/pull/178) · [2/6 simplecv](https://github.com/rerun-io/examples-monorepo/pull/179) · [3/6 mv-api](https://github.com/rerun-io/examples-monorepo/pull/180) · [4/6 geometry core](https://github.com/rerun-io/examples-monorepo/pull/181) · [5/6 catalog + Stage A/B](https://github.com/rerun-io/examples-monorepo/pull/182) · [6/6 refine + pipeline](https://github.com/rerun-io/examples-monorepo/pull/183). Merge in order, #185 first.

## What
- Rig-layout datasets (`/world/rig_NN/cam_MM`) are discovered through `simplecv.catalog_rig_layout` in one schema pass. The moving rig is skipped, cameras without pinhole + transform are dropped with a warning, partial calibration raises. Flat v1 layouts still work.
- `--keypoint-subset full` writes every COCO-133 joint; the default layer name follows the subset (`mvapi_coco133_full_v1`), so a full run no longer overwrites the upper-body layer.
- Per-stream codec comes from `VideoStream:codec` (the `--video-codec` flag is gone) and feeds the 0.37 `VideoFrameDecoder(codec=)`. TensorRT is the default tracker backend.

## Verified
ruff, pyrefly; mv-api tests 56. Ran on wildcap-mamma (H.264) and assembly101 (AV1) on the Spark at 36 frames/s with the 0.37 batch dataloader.

`catalog_prediction_layer.py` is 1,406 lines (1,383 on main); splitting out the viewer-validation half is a follow-up.

PR #183, opened 2026-09-03
exo-calib: refinement, report, blueprint, and the pipeline driver (stack 6/6)

pablo-rerun:

## What
- `refine` (Stage C+D): AHX post-processing → Kineo correspondences → robust triangulation against the init rig → Schur BA with centre priors → focal stage + one BA round → rotation-update guard → MoGe-2 metric rescale → `exocalib_refined`. Cameras without dataset calibration get the refined one on their base node.
- `report` (Stage E): SE(3) / Sim(3) errors against ground truth into `eval.json` (typed `EvalReport`), plus the `exocalib_align` viewer layer.
- `blueprint`: the exoego viewer layout, registered as the dataset default.
- `pipeline`: one task (`exocalib-pipeline`): Stage A, then Stage B + refine through the init rig and again through the refined rig, then the report.

## Results (rerun 0.37 catalog, DGX Spark, 2026-09-03)

| dataset | Sim3 cm | rot ° | SE3 cm | init SE3 cm / ° |
|---|---|---|---|---|
| assembly101 (8 cams) | 1.51 | 1.14 | 1.9 | 6.9 / 5.4 |
| hocap | 1.77 | 1.66 | 1.9 | 3.1 / 2.2 |
| epfl-smart-kitchen | 9.5–12.6 | 4.1–5.0 | 16–19 | 24.2 / 3.4 |
| wildcap-mamma hugging (vs MAMMA meta) | 3.27 | 0.47 | — | — |

Sim3 and rotation are stable run to run; SE3 carries the MoGe rescale noise. EPFL is guard-dominated (the rotation guard reverts four of five cameras) and swings between runs on the same code.

## Verified
ruff, pyrefly, vulture, 36 tests (beartype on); the full pipeline on the four segments above. Viewer screenshots under `/tmp/rerun-viewer-validation/exocalib-037/` on the Spark.

### DIFF
--- a/packages/exo-calib/exo_calib/apis/blueprint.py
+++ b/packages/exo-calib/exo_calib/apis/blueprint.py
@@ -1,0 +1,87 @@
+"""Write and register the exocalib viewer layout as the catalog dataset's default blueprint.
+
+Catalog datasets carry no layout of their own: the converters' base rrds embed a
+blueprint the catalog does not serve, and only the dataforge converters register a
+dataset default (a 2D grid for wildcap, whose bases have no calibration to place in
+3D). This stage registers the exoego layout — a 3D scene, the ego cameras tabbed
+beside it, the exo cameras as a strip below — plus the two visibility rules only a
+blueprint can express: Stage B's queryable ``…_raw`` record stays out of the 2D
+views, and an ego camera without calibration (no pose in any layer) stays out of
+the 3D scene. Everything else about how exocalib data looks — the init / refined
+frustum colours, class colours, radii — is logged with the data.
+
+Catalog partitions use the dataset id as their application id, and the id changes
+whenever the dataset is recreated, so rerun this tool after a dataset recreate.
+"""
+
+from dataclasses import dataclass, field
+from pathlib import Path
+
+import rerun.blueprint as rrb
+import tyro
+
+from exo_calib.catalog_io import RigLayout, StageConfig, StageContext, stage_context
+from exo_calib.kp2d_layer import kp2d_raw_entity
+from exo_calib.layer_io import pinhole_entity
+
+BLUEPRINT_FILENAME: str = "exocalib.rbl"
+"""Written under ``output_dir``; a ``-N`` suffix is added if it exists (the server caches file descriptors, so registered files are never overwritten)."""
+
+
+@dataclass
+class BlueprintConfig:
+    """Config for the blueprint writer; ``register`` sets the .rbl as the dataset's default blueprint."""
+
+    stage: tyro.conf.OmitArgPrefixes[StageConfig] = field(default_factory=StageConfig)
+    """Shared stage flags (catalog, dataset, segment, rigs, output, register); the CLI shows them unprefixed."""
+    loop_playback: bool = False
+    """Loop the whole timeline instead of parking on the last frame; for short standalone exports of a segment."""
+
+
+def main(config: BlueprintConfig) -> None:
+    """Recreate the exoego viewer layout, bound to the dataset's current id.
+
+    Mirrors ``simplecv.apis.view_exoego.create_container`` (the base rrd's own
+    blueprint, which the catalog does not serve). Not imported from simplecv
+    because that module pulls in dataset configs with deps outside this env.
+    """
+    context: StageContext = stage_context(config.stage)
+    layout: RigLayout = context.layout
+    exo_tabs: list[rrb.Tabs] = [
+        rrb.Tabs(rrb.Spatial2DView(origin=pinhole_entity(name), name=name, overrides={kp2d_raw_entity(name): rrb.EntityBehavior(visible=False)}))
+        for name in layout.exo_camera_names
+    ]
+    ego_tabs: list[rrb.Tabs] = [rrb.Tabs(rrb.Spatial2DView(origin=pinhole_entity(name), name=name)) for name in layout.ego_camera_names]
+    main_view: rrb.Horizontal = rrb.Horizontal(
+        contents=[
+            rrb.Spatial3DView(
+                origin="/",
+                name="3D View",
+                contents="/**",
+                spatial_information=rrb.SpatialInformation.from_fields(show_axes=True),
+                # An ego camera without dataset calibration has no pose in any layer, so keep it out of the 3D scene.
+                overrides={
+                    pinhole_entity(name): rrb.EntityBehavior(visible=False)
+                    for name in layout.ego_camera_names
+                    if name not in layout.calibrated_camera_names
+                },
+            ),
+            rrb.Vertical(contents=ego_tabs) if ego_tabs else rrb.Tabs(),
+        ],
+        column_shares=[4, 1],
+    )
+    container: rrb.Vertical = rrb.Vertical(main_view, rrb.Horizontal(contents=exo_tabs), row_shares=[4, 1])
+    # An explicit time panel opts out of ``collapse_panels``, so restate the collapsed state.
+    panels: list[rrb.TimePanel] = [rrb.TimePanel(loop_mode="All", state=rrb.PanelState.Collapsed)] if config.loop_playback else []
+    blueprint: rrb.Blueprint = rrb.Blueprint(container, *panels, collapse_panels=True)
+    config.stage.output_dir.mkdir(parents=True, exist_ok=True)
+    rbl_path: Path = config.stage.output_dir / BLUEPRINT_FILENAME
+    counter: int = 1
+    while rbl_path.exists():
+        rbl_path = config.stage.output_dir / f"{Path(BLUEPRINT_FILENAME).stem}-{counter}.rbl"
+        counter += 1
+    blueprint.save(str(context.dataset.id), str(rbl_path))
+    print(f"wrote {rbl_path} for application id {context.dataset.id}")
+    if config.stage.register:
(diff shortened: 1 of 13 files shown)


pablo-rerun:
Simplification, ablated on the three GT rigs and two MAMMA segments: refine uses every Stage B frame (no stride), the centre-prior sigma is `POSE_PRIOR_SIGMA_RATIO = 0.06` × the rig's median camera spacing, and Stage B caps at 1000 frames spread over the capture. Four tuning constants and one helper removed. GT unchanged within noise; MAMMA long_jump improved 27.6 cm / 3.2° → 9.4 cm / 0.9°.

pablo-rerun:
Thermo-nuclear review pass (2026-09-03), applied across the stack:
- Stage constants are the only copy of every tuned value; the geometry functions take them as required arguments (12 shadowing defaults removed, 2 of which disagreed with production).
- `StageConfig` + `stage_context()` replace five copies of connect / select segment / discover rig.
- `layer_io` keeps the contract; `kp2d_layer` (record round-trip) and `metric_scale` (MoGe rescale) are their own modules; the overlay lives in the Stage B tool.
- Canonical simplecv readers: `first_valid_value`, `read_camera_calibration`, `catalog_codec_name`, `read_rig_kinds`. `RefinementDiagnostics` is one dataclass for writer, reader and `eval.json`.
- Bug fix: the YOLOX re-anchor is time-based (`DETECTOR_REANCHOR_INTERVAL_S = 2.0`); the old `sample_idx % 60` misfired after the frame-budget sampler.
- Umeyama regains its ≥3-correspondence guard; a two-camera rig raises instead of producing a NaN prior.

Verified: package gates on linux-64 and aarch64; GT pipelines within noise (table in the description).

packages/exo-calib/exo_calib/apis/blueprint.py
```diff
@@ -0,0 +1,89 @@
+"""Write and register the exocalib viewer blueprint for the catalog dataset.
```
pablo-rerun: shouldn't we be getting this from dataforge? like doesn't dataforge already produce the right blueprint so having this feels extra?

packages/exo-calib/exo_calib/apis/pipeline.py
```diff
@@ -0,0 +1,39 @@
...
+
+    frame_index: int = 0
+    """Sample index of the synchronized frame Stage A calibrates from."""
+    batch_size: int = 32
+    """Largest inference batch the Stage B TensorRT engines are built for."""
+
+
+def main(config: PipelineConfig) -> None:
+    """Run every stage in order, registering each layer into the catalog."""
+    shared: dict[str, Any] = {field.name: getattr(config, field.name) for field in fields(StageConfig)}
```
pablo-rerun: not good to have any here, look through all of the pr stacks and please make sure we aren't using inprecise types unless theres no alternative

packages/exo-calib/exo_calib/metric_scale.py
```diff
@@ -0,0 +1,110 @@
...
+    ]
+    streams: ExoVideoStreams = open_exo_streams(dataset, segment_id, names)
+    moge: MoGeV2MetricPredictor = MoGeV2MetricPredictor(device="cuda")
+
+    # All surviving joints feed the estimator: body-only sampling has too few
+    # correspondences, and MoGe's body depth runs a few percent short.
+    observation_frames: Int64[ndarray, "n_obs"] = obs.point_frame_idx[obs.obs_point_idx]
+    group_medians: list[float] = []
+    n_samples: int = 0
+    for view in np.unique(obs.obs_view_idx):
```
pablo-rerun: lots of for loops here, can't we make thsi more vectorized? can you test it out making sure they're the same using something like hypothesis and then tell me the runtime/mem usage comparing using forloops like this v.s. doing things ina vectorized way and which way you would recommend

packages/exo-calib/exo_calib/apis/blueprint.py
pablo-rerun: Checked what the converters register. dataforge's `register` step does set a dataset default blueprint, but only for the datasets dataforge converts (wildcap, selfcap, robocap), and the wildcap one is a 2D camera grid with no 3D view, because those bases carry no calibration to place in 3D. assembly101, hocap and epfl-smart-kitchen were converted by the older egoexo-forge / simplecv path, which embeds a blueprint in each base rrd; the catalog does not serve embedded blueprints, so those datasets have no default at all until exo-calib registers one.

So the stage is not redundant yet, but it was doing too much. The init / refined frustum colours are now logged with the data (`Pinhole.color`, one static component per frustum, `VARIANT_COLOR` in `layer_io`), so the blueprint no longer carries colour overrides. What is left is what only a blueprint can express: the exoego layout (3D scene, ego tabs beside it, exo strip below) and two visibility rules (Stage B's queryable `…_raw` record stays out of the 2D views; an ego camera with no calibration, hence no pose in any layer, stays out of the 3D scene). The docstring says so.

The path to deleting the stage is for the converters to register a 3D layout default for calibrated datasets; then exo-calib only adds data. That belongs in dataforge / egoexo-forge, not in this stack.

packages/exo-calib/exo_calib/apis/pipeline.py
pablo-rerun: Done, and the sweep ran over every non-test file in the stack.

- `pipeline.py`: the `dict[str, Any]` is gone. The stage configs now compose `StageConfig` (`stage: tyro.conf.OmitArgPrefixes[StageConfig]`) instead of inheriting it, so the driver builds `Keypoints2dConfig(stage=config.stage, batch_size=…)` and passes `config.stage` to refine / report. The CLI flags are unchanged (`--catalog-url`, `--register`, … stay unprefixed).
- Also fixed by the sweep: `report.py`'s `dict[str, list]` is a typed `RigGeometryReport`; every bool mask is `Bool[ndarray, …]` instead of bare `ndarray`; the Stage B record reader has no `cast` and no bare `list` (a `BOX_SOURCE_BY_NAME` lookup narrows the Literal); `read_rig_kinds` narrows `RigKind` through a lookup instead of a `pyrefly: ignore`; the TorchCodec index ignore in `video_io` is a typed `get_frame_at`; the `object`/`Any` results of `first_valid_value` go through a new typed `first_valid_value_as(column, kind)` in simplecv; the blueprint dicts are typed; the mv-api codec reader lost its `Any`; the test doubles use `object`, `StoreEntry` and jaxtyping instead of `Any`/`ndarray`.
- Left, on purpose: `ba.py`'s `# pyrefly: ignore[missing-import]` (compiled `kornia_rs.k3d`, no stub); tomllib results typed `dict[str, Any]` in two tests (TOML is untyped); and the pre-existing items this stack did not add: 46 `Any` in mv-api's catalog seams (`dataset_entry`, `segment_view`, … — main had 60, the stack removed 14; the rest need Protocols so the test fakes still pass beartype, a follow-up together with the file split), the `Any`s in trtkit's ONNX Runtime wrapper, and three `cast`s in simplecv's metric exoego calibration.

Re-verified after these changes: package gates green (exo-calib 45 tests, mv-api 56, simplecv unchanged), assembly101 pipeline Sim3 1.51 cm / rot 1.14° / SE3 1.48 cm (before: 1.51 / 1.14 / 1.89, SE3 is the MoGe rescale swing), MAMMA hugging 3.21 cm / 0.53° (before 3.27 / 0.47°).

packages/exo-calib/exo_calib/metric_scale.py
pablo-rerun: Measured three shapes of the same computation (MoGe inference excluded, it is identical in all three): the loop as written, a **hybrid** that assigns every observation to its (view, frame) group in one pass and takes all group medians with one sort (depth still sampled per frame image), and a fully **stacked** version that indexes one `[groups, h, w]` depth tensor. hypothesis (300 random problems: 1–6 views, up to 2,000 observations, NaN points, odd/even group sizes) confirms all three return identical medians and sample counts. Realistic sizes, best of 7, tracemalloc peak:

| case | loop | hybrid | stacked |
|---|---|---|---|
| assembly101-like: 8 views, 89k obs, 1080p depth | 3.3 ms / 1.4 MiB | 1.3 ms / 2.8 MiB | 43.6 ms / 759 MiB |
| hocap-like: 8 views, 137k obs | 4.7 ms / 2.1 MiB | 0.7 ms / 4.3 MiB | 42.9 ms / 760 MiB |
| epfl-like: 5 views, 11k obs | 0.9 ms / 0.2 MiB | 0.2 ms / 0.3 MiB | 27.0 ms / 475 MiB |
| stress: 1,200 groups, 89k obs, 270p depth | 70 ms / 1.4 MiB | 63 ms / 15 MiB | 82 ms / 1.2 GiB |

Recommendation, applied: the hybrid. It is 2.5–6× faster than the loop at production sizes and reads better (one grouping pass, one `grouped_medians`, and the only remaining loop is the one MoGe forces: one forward pass per frame). The stacked version is the wrong trade: it is slower and needs every depth map resident (760 MiB at 1080p). In absolute terms none of this matters for wall time — the numpy part is a few milliseconds against 48 MoGe forward passes — so the change is for clarity, and the hypothesis test now lives in `tests/test_metric_scale.py` (the loop stays there as the reference).

pablo-rerun:
Landed on main as #211 (same content, rebased onto main after GitHub's stack tooling closed this PR when its base branch was deleted following the #178 merge).

PR #185, opened 2026-09-03
workspace: pin sibling packages with [tool.uv.sources]; guard the lock against PyPI namesakes

pablo-rerun:
## What
- Every package whose `[project].dependencies` names a sibling now pins it to the in-repo path with `[tool.uv.sources]`. Without the pin pixi resolves the bare name through uv as a registry requirement, and PyPI hosts unrelated `monopriors`, `simplecv` and `sam3` packages.
- `sapiens-coco133-pose` drops `mv-api` (nothing imports it); its envs had pulled the unrelated PyPI `monopriors` 0.1.0.
- `simplecv/tests/test_workspace_sources.py` fails if a sibling is unpinned or the lock ever resolves a workspace name from an index.
- README and AGENTS.md record the convention. `pixi.lock` regenerated; only the two sapiens envs differ from main.

## Verified
`pixi lock --check`, idempotent `pixi lock`, and the `tests` task of every changed package on the dl-server (linux-64), the Spark (aarch64) and the 3060 desktop.

Base of the exo-calib stack (#178–#183). Merge this first.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -70,7 +70,7 @@ packages/<name>/
 
 ## Adding a new package
 
-1. Create `packages/<name>/` with `pyproject.toml`, the source module, `tools/`, and `tests/` (structure above).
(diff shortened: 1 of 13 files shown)


pablo-rerun:
3060 desktop: the two monoprior DINOv2 golden tests fail identically on `main` there and pass on the 5090. Pre-existing tolerance, unrelated to this PR.

pablo-rerun:

PR #187, opened 2026-09-03
simplecv depth/mesh helpers; PromptDA teacher batch tool GPU cast + malloc_trim (stack 2/7)

pablo-rerun:
Stack **2/9** — the ZipDepth-PromptDA work re-cut into one focused stack (replaces #132 #133 #134 #129 #164 #165 #166 #167 #172 #173 #174 #175 #177; #131 is already merged). Review bottom-up.

|---|---|---|---:|

## simplecv depth/mesh helpers; PromptDA teacher batch tool on the GPU, malloc_trim per segment
**What** `simplecv.ops.depth.quantize_depth_m_to_mm` (torch, on-device) and `simplecv.ops.tsdf_depth_fuser.log_fused_mesh` (typed `mesh: o3d.geometry.TriangleMesh`, beside the fuser that produces the meshes; entity-first; keyword `recording`/`static`/`face_rendering`) replace inlined copies in view_polycam_data, monoprior polycam_inference and the prompt-da ARKitScenes tools.
The chosen-frame PromptDA batch tool (`prompt-da-arkitscenes`) casts depth to uint16 and downsizes the fusion RGB on the GPU before the D2H copy (22.4 → 6.3 MB/frame, 17.3 → 21.6 frames/s on the 5090; stored bytes and layer contract unchanged; A/B mesh vertex count identical) and returns Open3D's TSDF arena pages to the OS at each segment boundary (`gc.collect()` + `malloc_trim(0)`), which fixed the ~0.6 GB/segment host growth that OOM-killed the corpus batch at 142 GB RSS (RSS now 4–8 GB across 570 segments).
AGENTS.md gains the catalog-server rules from that incident: never kill Rerun by name, paseo's cgroup `KillMode`, tmux / systemd-run for long jobs.
**Gates** simplecv-dev 152 passed; prompt-da-dev 8 passed + deadcode; monoprior-dev lint/typecheck.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -109,6 +109,28 @@ layer. `arkitscenes-download-serve` uses `ulimit -n 524288`; restart the server
 to inherit it. This is a capacity workaround—the upstream fix is open-on-demand
 files or an LRU descriptor pool.
 
+### The shared catalog server (:51235) is in-memory — never kill Rerun by name
+
+`rerun server` keeps every registration in memory: any exit, even SIGTERM, loses
+the whole corpus and re-registering takes ~90 s warm / ~16 min cold. It has died to
+two things, both from agent sessions on this host:
+
+- **Name-based kills** (`pkill -x rerun`, `pkill rerun`, `killall rerun`,
+  `pkill -f rerun`) hit every process called `rerun`. Kill your own viewer by pid or
+  with a pattern that includes your port, e.g. `pkill -f "rerun --headless --port 9877"`,
+  and check `ss -ltnp | grep 51235` before touching anything named `rerun`. (The
+  server currently runs through a `rerun-catalog` symlink so its process name is not
+  `rerun`; that hides the symptom, it does not excuse a name-based kill.)
+- **paseo restarts.** `paseo.service` uses `KillMode=control-group`, so a restart —
+  including the one systemd triggers when the kernel OOM-kills any process in that
+  cgroup — SIGTERMs every agent session and everything they spawned.
+  `setsid`/`nohup`/`disown` do not leave the cgroup. Start long-lived servers and
+  batches with `tmux new-session` (lives in the SSH session scope; verify with
+  `cat /proc/<pid>/cgroup`) or `systemd-run --user`.
+
+Real fixes, not yet done: a persistent catalog store, and running the server under
+a supervisor instead of a shell.
+
 ## Testing Rerun builds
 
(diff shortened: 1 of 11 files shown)


packages/simplecv/simplecv/rerun_log_utils.py
```diff
@@ -507,6 +507,40 @@ def mesh_bounding_geometry(vertices: Float64[ndarray, "n 3"]) -> tuple[Float64[n
     return center, radius
 
 
+def log_fused_mesh(
+    entity_path: str,
+    mesh: Any,
```
pablo-rerun: we should be more specific here, this shouldn't just be Any. We should never use any unless theres absolutely no way around it.

AGENTS.md
```diff
@@ -109,6 +109,41 @@ layer. `arkitscenes-download-serve` uses `ulimit -n 524288`; restart the server
 to inherit it. This is a capacity workaround—the upstream fix is open-on-demand
 files or an LRU descriptor pool.
 
+### The catalog server is in-memory and dies to `pkill -x rerun`
```
pablo-rerun: this is good to have but too verbose, can you make it more concise please?

packages/simplecv/simplecv/rerun_log_utils.py
pablo-rerun: Fixed, and applied stack-wide. `log_fused_mesh` moved beside the Open3D fuser that produces these meshes (`simplecv.ops.tsdf_depth_fuser`) and is typed `mesh: o3d.geometry.TriangleMesh`; it could not stay in `rerun_log_utils` because 40 of the 66 envs that install simplecv have no open3d. I then audited every `Any` the whole stack adds (58 lines across 15 files) and removed all of them: typed pyserde configs/reports instead of `dict[str, Any]`, a `SegmentRow` dataclass for catalog rows, `runtime_checkable` Protocols for the instrumented dataset and the export model, a `TypedDict` for albumentations results, and validated `dict[str, Tensor]` narrowing of `torch.load` results. `git diff main -- '*.py' | grep '^+' | grep -w Any` is empty on every PR of the stack.

AGENTS.md
pablo-rerun: Condensed from 35 to 22 lines: same two rules (never kill Rerun by name; paseo's cgroup `KillMode` means long-lived jobs go through tmux or `systemd-run --user`), the incident dates and the 142 GB anecdote dropped, the "real fixes" kept as one sentence.

PR #188, opened 2026-09-03
monoprior: config-first depth-completion zoo, PromptDA torch/onnx/tensorrt (stack 3/7)

pablo-rerun:
Stack **3/9** — the ZipDepth-PromptDA work re-cut into one focused stack (replaces #132 #133 #134 #129 #164 #165 #166 #167 #172 #173 #174 #175 #177; #131 is already merged). Review bottom-up.

|---|---|---|---:|

## monoprior: config-first depth-completion zoo (PromptDA torch/onnx/tensorrt)
**What** Per-model Tyro configs with a `backend` subcommand (torch | onnx | tensorrt). PromptDA's ONNX/TensorRT export moves from prompt-da into `monopriors/models/depth_completion/prompt_da_export.py` beside the model; prompt-da's `trt_engine`/`trt_predictor` become compat shims over it. `base_completion_depth.py` owns the shared contracts once: binding names, `PROMPT_DEPTH_HW` (192×256), the `[0.1, 4.0]` m metric-depth window, `preprocess_completion_batch`, and a batched `Float32[Tensor, "b h w"]` metre-depth result (`CompletionDepthPrediction` stays as a compatibility wrapper). `stack_polycam_batch` lives beside `PolycamData` in simplecv so every Polycam tool builds model inputs the same way.
**Refactor only**: no new model. The registry is a plain module here; the next PR adds ZipDepth-PromptDA and the two-entry subcommand union.
**Review focus** `base_completion_depth.py` (the contract), `prompt_da.py` config/predictor, the export move.
**Gates** monoprior-dev 79 passed + deadcode; prompt-da-dev 8 passed + deadcode; simplecv-dev 155 passed.


packages/monoprior/monopriors/apis/promptda_polycam.py
```diff
@@ -98,19 +102,27 @@ def pda_polycam_inference(
...
         rr.set_time("frame_idx", sequence=frame_idx)
-        # convert image data to tensor
-        depth_pred: CompletionDepthPrediction = model(
-            rgb=polycam_data.rgb_hw3, prompt_depth=polycam_data.original_depth_hw
+        rgb_bhw3: UInt8[Tensor, "1 h w 3"] = torch.from_numpy(polycam_data.rgb_hw3).unsqueeze(0).cuda()
+        prompt_bhw: Float32[Tensor, "1 192 256"] = (
+            torch.from_numpy(polycam_data.original_depth_hw).unsqueeze(0).to(device="cuda", dtype=torch.float32) / 1000.0
         )
+        depth_bhw: Float32[Tensor, "1 h w"] = model(rgb_bhw3, prompt_bhw)
+        depth_pred_mm: UInt16[np.ndarray, "h w"] = (depth_bhw[0] * 1000.0).clamp(0.0, 65535.0).to(torch.uint16).cpu().numpy()
```
pablo-rerun: shouldnt' we use that recently added m -> mm depth util function you recently included in a previous pr?

packages/monoprior/monopriors/models/depth_completion/base_completion_depth.py
```diff
@@ -1,33 +1,141 @@
...
+        dtype=torch.float32,
+        memory_format=torch.contiguous_format,
+    ).div_(255.0)
+    if tuple(int(dim) for dim in image_b3hw.shape[-2:]) != image_hw:
+        image_b3hw = F.interpolate(image_b3hw, size=image_hw, mode="bilinear", antialias=True)
+    prompt_b1hw: Float32[Tensor, "b 1 192 256"] = rearrange(prompt_depth_bhw, "b h w -> b 1 h w")  # pyrefly: ignore  # bad-argument-type — einops stub false positive
+    return image_b3hw.contiguous(), prompt_b1hw.contiguous()
+
+
+def postprocess_completion_depth(
```
pablo-rerun: this name is a bit ambigous, should we be more specific? since we're basically just upsamping? I feel like this hids things. Would it be worth just inlining this? what do you think

packages/monoprior/monopriors/models/depth_completion/prompt_da_export.py
```diff
@@ -0,0 +1,100 @@
...
+class _FusionBreakerOutputs(nn.Module):
+    """Materialize prompt extrema to prevent a miscompiled TensorRT fusion."""
+
+    def __init__(self, model: nn.Module) -> None:
+        super().__init__()
+        self.model: nn.Module = model
+
+    def forward(self, image: Tensor, prompt_depth: Tensor) -> tuple[Tensor, Tensor, Tensor]:
+        """Return depth plus the prompt range fusion barriers."""
+        depth_b1hw: Tensor = self.model(image, prompt_depth)
```
pablo-rerun: any reason we aren't using jaxtyping here? in prompt da export? like it seems like these values should be specified? the shape + dtype? anywhere else in the code you didn't ofllow this where you should ahve?

packages/monoprior/monopriors/apis/promptda_polycam.py
pablo-rerun: Yes. Every hand-rolled `(depth * 1000).clamp(...).to(uint16)` the stack added now calls `simplecv.ops.depth.quantize_depth_m_to_mm` (this site, `prompt_da_polycam`, the Gradio UI, and the TRT Polycam comparison). One behaviour change worth knowing: the helper rounds to nearest instead of truncating, so stored millimetres can move by 1 mm.

packages/monoprior/monopriors/models/depth_completion/base_completion_depth.py
pablo-rerun: My take: keep one helper, fix the name, don't inline. It has two real callers with different runtimes (the predictor's `__call__`, and the teacher batch tool which drives the TensorRT runtime itself), and the policy it encodes (bilinear, `align_corners=False`, no antialias, always an owning tensor) moves edge metrics and the parity tests, so it should have one owner. Renamed to `upsample_depth_to_image` with that policy stated in the docstring; the `postprocess_depth` compat alias in `prompt_da.py` is gone and its one caller imports the real name. If the batch tool ever moves onto the predictor, it becomes single-caller and inlining is right then.

packages/monoprior/monopriors/models/depth_completion/prompt_da_export.py
pablo-rerun: No good reason — it was an omission, and not only here. I inventoried every array annotation the stack adds without dtype+shape (109 across 24 files, including the ONNX export wrappers' `forward` signatures and their tuple returns) and gave each a jaxtyping annotation; the export/parity tests still pass with beartype checking them at runtime. Related follow-up from your other comment: identifiers that repeat dtype or shape in the name (`_bhw3`, `_b1hw`, `_u8`) also violate the conventions; that rename pass (~60 names the stack introduces) is queued as a separate change so it does not hold up this review.

PR #190, opened 2026-09-03
zipdepth-catalog: streaming PromptDA pseudo-label lane, CUDA builders, holdout eval (stack 5/7)

pablo-rerun:
Stack **5/9** — the ZipDepth-PromptDA work re-cut into one focused stack (replaces #132 #133 #134 #129 #164 #165 #166 #167 #172 #173 #174 #175 #177; #131 is already merged). Review bottom-up.

|---|---|---|---:|

## zipdepth-catalog: streaming PromptDA pseudo-label lane, CUDA sample builders, holdout eval
**What** `zipdepth-catalog` / `zipdepth-catalog-dev` envs (torchcodec NVDEC, imagecodecs, ffmpeg ≥ 8.1 b902, nvidia-ml-py, editable arkitscenes-download) and `zipdepth/catalog/`, which turns the arkitscenes-v2 `promptda` layer into training samples without leaving the GPU:
- `targets` — the numerical contract (uint16 metric PNG → metres; validity from the shared `[0.1, 4.0]` m window), exact label geometry (nearest-exact depth, antialiased bilinear RGB, flip after resize — the eval transform shares the torch resize helper), resize-only training transform.
- `png` — PNG Up-filter reversal as an int32 cumsum on the GPU. `builders` — CPU and CUDA sample builders that agree bit-for-bit (tests), running libdeflate → unfilter → span filter → resize/flip on a producer-owned stream.
- `dataset` — one datafusion query per segment at the chosen-packet timestamps, zero-copy Arrow blob views, an NVDEC decoder over the segment's AV1 packets, N whole-segment producer threads → bounded shuffle buffer, rank sharding. `segments` — one connection, projected segment table, the deterministic holdout split shared by train and eval. `stats`, `threaded`.
- `zipdepth-catalog-throughput` (frames/s, per-stage ms); `zipdepth-eval-catalog` (unaligned metric AbsRel/δ1/MAE vs held-out PromptDA labels; edge/flat-stratified MAE and edge-weighted MAE on the teacher gradient stencil, `zipdepth.evaluation.edge_metrics`; a zero-parameter prompt-upsample baseline).
- simplecv `open_segment_decoder` drops the client-side `.sort()` (a SortExec re-materialised the blob columns, ~4× the query wall) behind an order guard that fails loudly.
**Review focus** `dataset.py` and `builders.py` (the producer hot path); `targets.py` geometry.
**Lock** adds the two envs; every other environment unchanged (byte-identical to the tested workspace lock).
**Gates** zipdepth-catalog-dev 54 passed + lint/typecheck/deadcode; simplecv-dev 155 passed.


packages/zipdepth/zipdepth/catalog/dataset.py
```diff
@@ -0,0 +1,400 @@
...
+            views.append(all_bytes_n[start:end])
+    return views
+
+
+def promptda_blob_views(column: pa.ChunkedArray) -> list[UInt8[ndarray, "n_bytes"]]:
+    """Return zero-copy PromptDA blob views, retaining the established public helper."""
+    return component_u8_views(column, PROMPTDA_BLOB_COLUMN)
+
+
+class CatalogPromptDepthDataset(IterableDataset[dict[str, Tensor]]):
```
pablo-rerun, line 141: is how we're doing this the idiomatic way of doing it in 0.37? This feels a bit insane to me but I might be wrong

https://rerun.io/docs/howto/train/dataloader

take a look here to be certain

pablo-rerun: You're right that it is not the idiomatic path. Idiomatic 0.37 is `RerunIterableDataset` + `Field(column, decode=ColumnDecoder)`; our own prompt-da stream already does that with `SegmentNvdecDecoder` as a custom decoder. This module owns its query, shuffle buffer, producer threads and rank sharding because the pre-0.37 dataloader (per-sample decode, Arrow→Python copies) fed training at ~17 frames/s against a GPU that needs ~150; the hand-rolled path measured 151. 0.37's batch decode contract (`decode(FieldBatch, requests)`) removes the reason: the GPU PNG path can be a custom `ColumnDecoder` and the rest goes back to the dataloader. The Arrow buffer walk you flagged is what any zero-copy decoder does with a `FieldBatch` (it hands you a `pa.Array`, no per-row bytes helper), so it survives inside the decoder. I'm measuring the idiomatic version side by side right now (throughput probe + the 1,500-step instrumented loop, plus a bit-identical-sample test) and auditing every catalog query in the stack against the round-trip patterns; result lands as a follow-up commit to this PR with the numbers.

pablo-rerun: Measured (report on pablo-dl-server: `/tmp/zdp-recut/apply/W8-REPORT.md`). An idiomatic `RerunIterableDataset` path now exists on a branch, reproduces the current samples **bit-identically** (live one-segment equivalence test), and reuses `SegmentNvdecDecoder` plus a batched CUDA PNG decoder as `ColumnDecoder`s. Paired 1,500-step training runs on the same 8 segments, RTX 5090:

| loader | median data wait | median throughput | GPU |
|---|---:|---:|---:|
| current (8 producers) | 7.7 ms | 153.7 frames/s | 84% |
| Rerun 0.37, fetch block 512 | 917.9 ms | 8.4 frames/s | stalls dominate |

Two API gaps cause the 18× gap, not our decoders: (1) a duration index requires `FixedRateSampling`; there is no public constructor for the exact irregular chosen-packet timestamps, so a private `SampleIndex` bridge must discover them corpus-wide (four concurrent 8-segment queries; the full 2,005-segment train split did not emit a first batch in 57 min); (2) the dataloader retains every field's native decoded output for a whole fetch block before emitting, so the depth-blob stage runs at 100–150 ms/frame against 12 ms in the threaded loader. Fetch block size, parallel inflation and shuffle-buffer changes did not move it.

Recommendation, as W8 also concluded: keep the threaded loader as the default, ship the Rerun path as an executable reference behind `--dataloader rerun` (it is the upstream target), and file the two API asks upstream (exact-timestamp sample index; a pre-retention decode/collate hook). The query audit found and fixed four PromptDA sites missing projections or filters (segment-table collects 6.4 → 0.9 MB; a chosen-frame query without a contents filter); everything in this module was already `filter_contents` + projection before `reader`, and `fill_latest_at` is required (without it every prompt/confidence cell is null).

PR #194, opened 2026-09-03
chore: drop the stale einops pyrefly suppressions

pablo-rerun:
Every `# pyrefly: ignore  # bad-argument-type — einops stub false positive` on a `rearrange` call is stale: with the einops the dev environments resolve (0.8.2), pyrefly 1.1.1 (locked) and 1.2.0 report **zero errors** on these lines with the suppressions removed (26 sites in 13 files on main). Comment-only change.

**Gates on this branch.** monoprior-dev lint/typecheck/tests green (77 passed; the pre-existing hypothesis failure in `tests/test_g3t_model.py::test_confidence_filter_matches_linear_percentile_mask` deselected, see below); prompt-da-dev and prompt-da-stream-dev green; gauss-surf-dev lint green. gauss-surf-dev typecheck and tests are red **on main itself** in this worktree (main:  INFO 2 errors (1 suppressed, 3 warnings not shown); 8 failed, 97 passed, 2 skipped in 5.66s) — unrelated to this change: the typecheck errors are the `extra_signals` kwargs of the pinned gsplat fork (fixed by the pyrefly baseline in #186), the test failures are the same eight on main.

**Why the suppressions existed.** The error shape reproduces only against einops 0.9.0.dev0, whose new typing (einops PR #430) bounds the `Tensor` TypeVar to a Protocol that leaves `__getitem__` untyped; pyrefly then rejects plain `torch.Tensor` (jaxtyping is not involved, its annotations erase to `Tensor`). The lock carries that dev release in exactly one env, `mv-api-catalog-register-mac`, the only feature with `einops = ">=0.8.0"` instead of `<0.9` — follow-up: tighten that pin. No einops or pyrefly issue tracks this; a local stub was evaluated and rejected as maintenance for a prerelease we do not use.

The ZipDepth-PromptDA stack (#186 → #192) drops the same comments in the files it touches, so the overlap merges cleanly either way. Notes with the version matrix: `/tmp/zdp-recut/einops/REPORT.md` on pablo-dl-server.

Unrelated bug found on the way: `test_confidence_filter_matches_linear_percentile_mask` fails on main once hypothesis finds a counterexample (NumPy's linear percentile vs the order-statistic mask in `multiview_model.py`); it deserves its own fix.

### DIFF
--- a/packages/gauss-surf/src/gauss_surf/apis/frame_selection.py
+++ b/packages/gauss-surf/src/gauss_surf/apis/frame_selection.py
@@ -99,7 +99,7 @@ def _score_camera(
     with torch.inference_mode():
         for frame_index in range(len(timestamps_n)):
             frame_chw: UInt8[Tensor, "3 h w"] = decoder.get_frame_at(frame_index).data
-            frame_hwc: UInt8[Tensor, "h w 3"] = rearrange(frame_chw, "c h w -> h w c")  # pyrefly: ignore  # bad-argument-type — einops stub
+            frame_hwc: UInt8[Tensor, "h w 3"] = rearrange(frame_chw, "c h w -> h w c")
             pending_frames_hwc.append(frame_hwc)
             if len(pending_frames_hwc) < batch_size:
                 continue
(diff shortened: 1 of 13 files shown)


pablo-rerun:
Superseded: the stack (#186–#204) removed every stale einops suppression in the files it touched, and after the merge `git grep 'einops stub'` on main returns nothing, so there is nothing left for this PR to do. The two follow-ups it carried stay valid: tighten the one `einops = ">=0.8.0"` spec (mv-api mac register env) to `<0.9`, and the pre-existing hypothesis counterexample in `tests/test_g3t_model.py`.

PR #216, opened 2026-09-04
zipdepth: ultrawide training lane (padded wide prompt, 50/50 sampler, footprint-split eval)

pablo-rerun:
Adds a rectified-ultrawide training lane to the ZipDepth-PromptDA catalog
pipeline, off by default. Full design, measurements and recipe:
`packages/zipdepth/docs/ultrawide.md`.

Stacked with #215 (`zipdepth/uw-1-range`, `range_margin_m`), which this branch
does **not** depend on in code — the trainer and loss are unchanged. Base first,
order as you like.

## Design

**`TrainCatalogConfig.cameras: "wide" | "ultrawide" | "both" = "wide"`**, plus
`ultrawide_min_valid_fraction=0.7`, `ultrawide_valid_erosion_px=1`,
`ultrawide_prompt_scale=1/2.1`. Same fields on the throughput probe, the smoke,
and the evaluation.

**Padded prompt, not reprojection.** The rectified ultrawide sees **2.1x** the
wide field of view on both axes, so the wide ARKit LiDAR prompt only covers its
centre. The oriented prompt is nearest-exact resampled to
`round(256/2.1) x round(192/2.1) = 122 x 91` and centred in a **zeroed** 192x256
canvas; zero already reads as "no prompt" through the model's `[0.1, 4.0] m`
gate. Reprojecting buys nothing: the decoder bilinearly resamples the prompt to
its finest H/4 level, which at 768x1024 input *is* 192x256, so a full-frame
canvas would be downsampled back to the same 122x91 block — and the ARKit map is
itself a few hundred LiDAR returns smeared over 256x192. Mirroring is applied to
the block before padding, so a horizontal flip stays exact for any canvas margin.

**The 2.1 constant**, measured from the catalog calibration over 24 segments
spanning both stored orientations, as the per-axis ratio of
`resolution / focal_length`:

```
fov ratio x: mean 2.1087  sd 0.0152  min 2.0688  max 2.1383
fov ratio y: mean 2.1087  sd 0.0152  min 2.0688  max 2.1383
principal-point offset from centre: <= 0.66% of width, <= 0.69% of height
```

**Prompt timing.** Ultrawide rows sit at their own 10 Hz selection, disjoint from
the wide chosen frames (on `47115416` only 1 of 651 timestamps coincides). Each
ultrawide frame takes the nearest ARKit map at or before its timestamp;
measured staleness on `47115416` and `42444511` is `max 16.1 ms, mean 7.0 ms`,
bounded by the 60 Hz ARKit interval.

**Mask policy.** `target_valid = 0.1 <= d <= 4.0`, then binary erosion of 1 px
(max pool over the inverted mask, one kernel for CPU and CUDA); a frame is
dropped when the valid fraction *before* erosion is under 0.7. Sampled valid
fractions: `47115416` mean 0.914 / p10 0.866 (95% of frames >= 0.7),
`42444511` mean 0.958 / p10 0.901 (98% >= 0.7). The existing flat-frame filter
is unchanged. Evaluation defaults the drop threshold to 0.0 — dropping sparse
frames is a training choice, not a property of the benchmark.

**50/50 sampler.** With `cameras="both"` each producer yields the segment's wide
chosen frames plus a seeded random ultrawide subsample **of equal size**,
round-robin interleaved, after which the existing shuffle buffer mixes them. The
subsample seed is the segment seed (run seed, epoch, rank, segment), so every
pass draws a fresh subset; the ultrawide augmentation seed stream is offset from
the wide one. A segment lacking the `ultrawide_depth` layer (1 of 2,025) falls
back to **wide-only** under `both`; a segment with no wide chosen frames is
skipped entirely under `both`, so its 10 Hz ultrawide track cannot skew the mix.

## Robustness (review round 2)

**Null payload rows can no longer kill a producer.** The ultrawide track has its
own `drop_leading` and its own 10 Hz selection, so a chosen frame can precede the
first ARKit lowres depth row; latest-at cannot fill a column that has not been
logged yet, and `component_bytes_views` rejects a null outer row. The lane now
drops any chosen row whose RGB, depth, prompt, or confidence cell is null and
counts it in `skipped_missing_payload_frames`. Filtered client-side rather than
in the datafusion chain: a null row carries no payload bytes, so this costs the
same transfer as a server-side filter and still yields the count in one query.

**Preflight over all 2,024 segments with the layer** (1,449,709 ultrawide chosen
frames, 716.3 per segment):

```
rgb_mismatch               0   ultrawide rgb timestamps != ultrawide depth timestamps
rgb_missing                0   chosen frames with no rgb row
before_first_prompt        0   chosen frames preceding the first ARKit lowres depth row
before_first_conf          0   chosen frames preceding the first ARKit confidence row
segments with any problem  0
```

So the hazard is **latent, not active**: as registered today nothing is missing,
and ultrawide RGB and depth share timestamps exactly in every segment. The drop
is a guard against a future re-registration, not a workaround for current data.

**Orientation is asserted, not assumed.** The layer writer
(`gauss_surf/apis/ultrawide_depth_batch.py`, PR #214) reads `image_wh` from the
calibration and has no orientation guard of its own, and 813 of the 2,024
segments store 480x640 portrait. Both builders now assert that an ultrawide frame
is landscape after its quarter turns and fail loudly otherwise. A new portrait
test (`quarter_turns=1`, mirroring `test_catalog_dataset.py:139`) checks that the
image, the target and the placed prompt come out upright and aligned, and a
parametrized test checks the guard fires for `quarter_turns` 0 and 2 on a stored
portrait frame.

**Counters surfaced.** `skipped_flat_frames`, `skipped_low_valid_frames` and
`skipped_missing_payload_frames` now appear in the training log line and the
TensorBoard `io/*` scalars alongside `skipped_frames`, and in the throughput
probe's summary.

Nits also fixed: `erode_valid` docstring now says `max_pool2d` pads with `-inf`
(never winning over a 0/1 mask) rather than "zero padding"; the dead
`perf_counter()` before the wide decode loop is gone; `_ordered_times` is renamed
`_require_ordered_rows` for its validate-only use on the ultrawide path.

Deferred by agreement: the parallax measurement (M2), and carrying the pass index
through resume so a resumed run does not re-draw the same ultrawide subsets (M3)
— M3 is recorded as a known limitation in the doc.

## Wide lane is bit-identical

`cameras="wide"` builds no ultrawide policy at all, the builders never consult a
policy they were not given, and `_resolve_placement` rejects an ultrawide sample
from a builder without one.
`test_wide_samples_are_bit_identical_with_and_without_an_ultrawide_policy` hands
one builder a policy that would reject every frame and erode 8 px, then asserts
every tensor of the wide sample is unchanged.

## Smoke (dl-server, RTX 5090)

`tools/train_catalog_smoke.py --cameras both --segment-ids 47115416 42444511`.
One fixed probe set **per camera** — a mixed set would hide a camera whose loss
rose behind another camera's larger fall — and the gate must fall for both:

```
loss gate passed on the fixed wide probe set:      released=0.760929, trained=0.540120  (-29.0%)
loss gate passed on the fixed ultrawide probe set: released=0.672580, trained=0.495493  (-26.3%)
catalog smoke passed; outputs in data/smoke/catalog-runs/latest
Elapsed (wall clock) time: 2:07.60
```

All three legs (120-step train, 4-step resume, 1-process torchrun DDP) ran with
`--cameras both`. **No `init_checkpoint`:** the smoke's legs train the bare
`ZipDepth-base` SSI lane, so they start from the released weights as they always
have. `data/checkpoints/zdpda-v4/final_model.pth` does exist on the host but is a
prompted *metric* checkpoint and is not loadable by that leg.

## Throughput (dl-server, 8 producers, batch 8, 768x1024, CUDA builder)

| cameras | segments | frames/s | segment_query | video_decode | jpeg_decode | blob_decode | augment |
| --- | --- | --- | --- | --- | --- | --- | --- |
| wide | 8 | 173.9 | 11.92 | 5.53 | 0.00 | 4.00 | 0.52 |
| both | 8 | 344.2 | 5.00 | 3.64 | 0.39 | 3.29 | 1.14 |
| ultrawide | 8 | 681.3 | 1.73 | 0.00 | 0.75 | 2.50 | 2.24 |
| both | 16 mixed-orientation, 60 s | 296.0 | 14.80 | 3.35 | 0.44 | 3.44 | 1.01 |

(stage columns ms/frame; the 8-segment rows are one full pass each.) The
16-segment mixed-orientation run built 18,109 samples with **0** missing-payload
drops and no orientation failures — 195 frames dropped by the valid-fraction gate
and 799 by the flat-frame filter.

The ultrawide lane is much cheaper than the wide one — no NVDEC, a 640x480 JPEG
instead of a 1920x1440 video frame — so mixing the cameras **doubles** loader
throughput instead of costing anything, and stays well ahead of the ~150
frames/s the trainer consumes. JPEG decode is the new stage and is not the
bottleneck.

**CPU JPEG decode, deliberately.** Measured here: nvjpeg 0.171 ms/frame batched
versus 0.566 ms/frame on the CPU — but nvjpeg's output differs from
libjpeg-turbo's by up to 27/255 per channel, which would break the CPU/CUDA
builder parity the probe and evaluation paths depend on, and torchvision's
nvjpeg handle is shared across producer threads. 0.4 ms/frame is not worth
either risk while the loader runs at twice the trainer's rate.

## Evaluation

`--cameras` also on `eval_catalog`. Ultrawide frames are scored whole, **inside**
the prompt footprint (the centred `[200:564, 268:756]` box of the 768x1024
output, derived from the placement), and **outside** it, with the existing
`prompt-upsample` floor computed for the inside region. Wide metrics and their
edge-stratified block are untouched.

Sanity run, `zdpda-v4` (wide-only training) on 2 segments at stride 120 — the
"before" picture the fine-tune has to move:

```
overall metric:    AbsRel=0.014593 delta1=0.992969 MAE=0.028748m frames=11   (wide, unchanged)
overall ultrawide: whole AbsRel=0.589050 MAE=0.751074m
                 | inside AbsRel=0.037448 MAE=0.072088m (prompt-upsample floor AbsRel=0.038276)
                 | outside AbsRel=0.760235 MAE=0.953446m delta1=0.258906
```

Inside the footprint the wide-only checkpoint sits on the upsample floor and
outside it has essentially nothing, which is exactly the gap the lane targets.

## Tests

`pixi run -e zipdepth-catalog-dev {lint,typecheck,deadcode,tests}` on
`pablo-dl-server`:

```
lint      ruff  All checks passed!
typecheck pyrefly  0 errors
deadcode  vulture  clean (no new ignore_names needed)
tests     131 passed, 2 skipped
```

Tests run with `--ignore=tests/test_infer_rerun_pixels.py`: that module fails to
import on `main` (73592b3) too, with `ModuleNotFoundError: No module named
'omegaconf'` from `monopriors.models.stereo_depth.fast_foundationstereo`. It is
**pre-existing and unrelated** to this branch — verified by collecting on the
pristine main checkout.

New `tests/test_catalog_ultrawide.py` (21 tests) covers placement geometry and
the 2.1 constant, zero-padding and validity, flip exactness on the block, the
erosion kernel including CPU/CUDA parity and border handling, the valid-fraction
drop and its separate counter, the seeded subsampler, `interleave`, the
footprint box, wide bit-identity, portrait orientation and the landscape guard,
and CPU/CUDA builder parity for the placed prompt (exact masks; prompt depth to
1 float32 ULP, which is the device's divide rounding).

Not merging — ordering is yours.

### DIFF
--- a/packages/monoprior/monopriors/models/depth_completion/zipdepth_prompt.py
+++ b/packages/monoprior/monopriors/models/depth_completion/zipdepth_prompt.py
@@ -3,9 +3,16 @@
 The wrapper keeps the vendored ZipDepth-base tensors unchanged, injects a
 normalized metric prompt into all four decoder stages, and returns depth in the
 same units as the prompt (metres for the public API).
+
+The head can only reach depths inside the range it is rescaled into, which by
+default is the prompt's own ``[min, max]``. ``ZipDepthPrompt(range_margin=...)``
+widens that range by a fraction of the prompt span on each side, which ultrawide
+captures need: their prompt covers the image centre, so the periphery is often
+nearer or farther than anything the prompt saw.
 """
 
 from dataclasses import dataclass, field
+from math import isfinite
 from pathlib import Path
 from typing import Self
 
@@ -33,7 +40,11 @@ from monopriors.models.depth_completion.base_completion_depth import (
     PROMPT_INPUT_NAME,
     BaseCompletionPredictor,
 )
-from monopriors.models.zipdepth_checkpoint import load_zipdepth_state_dict
+from monopriors.models.zipdepth_checkpoint import (
+    load_zipdepth_checkpoint,
+    range_margin_from_checkpoint,
+    state_dict_from_checkpoint,
+)
 from monopriors.third_party.zipdepth.architecture import EncoderOutput, FeaturePyramid, ZipDepth, create_model
 from monopriors.third_party.zipdepth.model_utils import StateDict
 
@@ -82,15 +93,36 @@ def _inject_prompt(
 
 
 class ZipDepthPrompt(nn.Module):
-    """ZipDepth-base with PromptDA-style multi-scale metric conditioning."""
+    """ZipDepth-base with PromptDA-style multi-scale metric conditioning.
 
-    def __init__(self, backbone: ZipDepth | None = None) -> None:
+    The head is a sigmoid rescaled into a per-image depth range derived from the
+    prompt. ``range_margin`` widens that output range symmetrically, so predicted
+    depth is no longer trapped inside the prompt's own ``[min, max]``. The prompt
+    *normalization* fed to the four decoder injection branches deliberately keeps
+    the original ``[min, max]``: the prompt itself does not change, only the range
+    the head may reach, and holding the conditioning fixed keeps a margin change
+    from shifting the features a trained model already relies on.
+    """
+
+    def __init__(self, backbone: ZipDepth | None = None, *, range_margin: float = 0.0) -> None:
+        """Build the prompted wrapper.
+
+        Args:
+            backbone: ZipDepth-base network; ``None`` creates a fresh one.
+            range_margin: Fraction of the prompt span by which the head's output
+                range is widened past the prompt's own minimum and maximum. Zero
+                reproduces the prompt-bounded head exactly, emitting none of the
+                widening operations.
+        """
         super().__init__()
         self.backbone: ZipDepth = create_model(variant="base") if backbone is None else backbone
         if self.backbone.variant != "base":
             raise ValueError(f"ZipDepthPrompt requires the base variant, got {self.backbone.variant!r}")
         if not self.backbone.decoder.convex_up.use_unfold:
             raise ValueError("ZipDepthPrompt currently requires ZipDepth's unfold-based convex head")
+        if not isfinite(range_margin) or range_margin < 0.0:
+            raise ValueError(f"range_margin must be finite and non-negative, got {range_margin}")
+        self.range_margin: float = range_margin
         self.prompt_branches: nn.ModuleList = nn.ModuleList(
             [_make_prompt_branch(channels) for channels in (288, 192, 144, 96)]
         )
@@ -130,10 +162,15 @@ class ZipDepthPrompt(nn.Module):
         Float32[Tensor, "b 1 1 1"],
         Float32[Tensor, "b 1 1 1"],
     ]:
-        """Predict metric depth and expose the prompt range for TRT export.
+        """Predict metric depth and expose the head's output range for TRT export.
 
         The two scalar outputs force TensorRT to materialize the input-derived
         reductions. Normal callers use :meth:`forward` and receive depth only.
+
+        Returns:
+            Metric depth, and the minimum and maximum the head can reach. Those
+            two already include ``range_margin``, so ``sigmoid(logits) * (max -
+            min) + min`` reproduces depth from captured logits.
         """
         if image.dtype == torch.uint8:
             image_bchw: Float[Tensor, "b 3 h w"] = image.to(dtype=self.backbone.mean.dtype) / 255.0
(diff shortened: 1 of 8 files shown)


pablo-rerun:
Result from the lane this PR builds, for the record.

**The ultrawide fine-tune it enables plateaued at 0.234 outside-footprint AbsRel** (`zdpda-uw-v1`, down from
v4's 0.708), and it cost the wide lane 27% (0.0136 → 0.0173). Inside the footprint it beat the bilinear
prompt-upsample floor by only 5.6% (0.0960 vs 0.1017). Mid-training checkpoints at 90k and 120k–130k steps were
already at the final number, so this is a plateau rather than an unfinished run.

**The pre-fill path reaches 0.141 outside on `zdpda-v4` with no fine-tuning at all** — MoGe-2 metric depth over
the same rectified ultrawide RGB, rescaled by the median LiDAR-to-MoGe ratio over the footprint, with the real
LiDAR block pasted back inside. δ1 outside 0.31 → 0.89, at ~30–36 ms/frame on a 5090. Running the fine-tuned
uw-v1 with that same prompt gives 0.1438, slightly *worse* than untouched v4: once the prompt is filled the
ultrawide fine-tune contributes nothing.

None of that argues against this PR. **The lane and the footprint-split evaluation are exactly what made the
finding possible** — the oracle prompt (0.050), the wide-camera coverage sweep (inside flat at 0.013–0.017,
outside 0.16–0.25 at ultrawide-sized crop) and the seven-condition pre-fill comparison are all eval-only
experiments run through this machinery, and together they are what identified prompt coverage — not the output
range cap, not the prompt placement, not target noise — as the binding constraint.

Full write-up, tables and pixel evidence:

PR #217, opened 2026-09-04
zipdepth: uw-v1 integration (range margin + ultrawide lane)

pablo-rerun:
Integration branch for the `zdpda-uw-v1` training run. This is the merge of
#215 (`zipdepth/uw-1-range` @ 2b99705) and #216 (`zipdepth/uw-2-lane` @ 566640f)
onto `main`, and carries **no new work of its own** beyond conflict resolution.

It exists so the long fine-tune can run against both features at once while the
two PRs are still in review. It will be rebased once #215 and #216 land, and
closed rather than merged if they land cleanly on their own.

## Conflict resolution

The only conflict was in `packages/zipdepth/zipdepth/apis/train_catalog.py`, in
the validation block at the top of `main()`, where both sides added adjacent
validation. The resolution keeps both, in order: the `range_margin_m`
finite/non-negative and metric-lane checks from #215, then the `cameras`
selection and Rerun-dataloader checks plus the `UltrawidePolicy` construction
from #216. Everything else auto-merged. The second merge (#216 @ 566640f, the
null-payload / orientation-assert / skip-counter fixes) was conflict-free.

Post-merge invariants checked:
- `TrainCatalogConfig` carries both `range_margin_m` and the `cameras` /
  `ultrawide_*` group.
- `ResolvedTrainCatalogConfig` flattens `TrainCatalogConfig`, so
  `resolved_config.json` records both.
- The resume margin guard (`_restore_resume_state`) and the init-time
  `OVERRIDE:` notice are intact.
- The checkpoint records `range_margin_m`, and the trainer still receives it.

## Gate results, and why uw-v1 trains at `range_margin_m = 0.0`

A 600-step real fine-tune from `zdpda-v4` was run as a gate before committing to
the 140k-step run. All numbers below use the same 4 holdout segments,
`--frame-stride 60`, 51 frames, `--cameras both`, evaluated with
`tools/eval_catalog.py`.

| checkpoint | steps | margin | lr | wide AbsRel | wide δ1 | wide MAE | uw whole | uw inside | uw outside | uw δ1 |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| `zdpda-v4` (baseline) | — | 0.0 | — | **0.012690** | 0.9942 | 0.0243 m | 0.7707 | 0.1478 | **0.9577** | 0.2528 |
| gate, as originally specified | 600 | 3.9 | 1e-4 | **0.1937** | 0.5970 | 0.3602 m | 0.4810 | 0.2745 | 0.5445 | 0.3692 |
| control, margin 0 | 600 | **0.0** | 1e-4 | **0.015466** | 0.9922 | 0.0292 m | 0.3445 | 0.1431 | **0.4072** | 0.5534 |
| control, margin 3.9 at v4 lr | 600 | 3.9 | 1e-5 | 0.2158 | 0.5448 | 0.4203 m | 0.3343 | 0.2811 | 0.3515 | 0.4532 |

The gate's pass criterion was "ultrawide outside-footprint AbsRel clearly below
v4's, and wide AbsRel not worse than v4 by more than 0.002". At
`range_margin_m = 3.9` the ultrawide half passes and the wide half fails by two
orders of magnitude.

**The cause is the margin, not the ultrawide data and not the learning rate.**
The margin-0 control — same `--cameras both`, same `fast` learning rate, same 600
steps — holds wide at 0.015466 (only +0.0028 over v4) while still delivering most
of the outside-footprint win (0.4072 against v4's 0.9577, δ1 0.5534 against
0.2528), and brings ultrawide inside-footprint AbsRel to 0.1431, just below the
zero-parameter prompt-upsample floor of 0.1540. Margin 3.9 breaks the wide lane
at **both** learning rates tested.

Mechanism: `zdpda-v4`'s weights were trained with the head's sigmoid spanning
`[prompt_min, prompt_max]`. A 3.9 m margin re-spans it to `[0.1, 4.0]` on every
frame, so an unchanged activation now decodes to a different depth — the model's
operating point moves and the mapping has to be relearned (step-1 L1 is 0.857 m).
`docs/ultrawide.md` argues the cost of the wider sigmoid "works out to a few
millimetres" of quantization precision; that reasoning accounts for output
resolution but not for the operating-point shift, and the measurement contradicts
it. Supporting evidence: the affine-aligned diagnostic for the margin-3.9 gate is
0.0740 while its raw metric AbsRel is 0.19–0.28, i.e. structure survives and it is
scale and offset that are wrong.

**uw-v1 therefore trains with `--range-margin-m 0.0`.** Whether a widened output
range helps once the model has had time to relearn the mapping is a separate
question, deferred to a uw-v2 experiment with a clean multi-thousand-step probe.

One measurement caveat, chased down and resolved: the margin-3.9 checkpoint does
not evaluate reproducibly (three runs gave 0.2832, 0.1937, 0.1937), whereas
`zdpda-v4` and the margin-0 control both repeat **bit-identically** across runs.
The harness is deterministic; the instability belongs to the margin-3.9 model.
The table quotes the margin-3.9 gate's best of three, and the conclusion holds at
either value.

### DIFF
--- a/packages/monoprior/monopriors/models/depth_completion/zipdepth_prompt.py
+++ b/packages/monoprior/monopriors/models/depth_completion/zipdepth_prompt.py
@@ -3,9 +3,16 @@
 The wrapper keeps the vendored ZipDepth-base tensors unchanged, injects a
 normalized metric prompt into all four decoder stages, and returns depth in the
 same units as the prompt (metres for the public API).
+
+The head can only reach depths inside the range it is rescaled into, which by
+default is the prompt's own ``[min, max]``. ``ZipDepthPrompt(range_margin=...)``
+widens that range by a fraction of the prompt span on each side, which ultrawide
+captures need: their prompt covers the image centre, so the periphery is often
+nearer or farther than anything the prompt saw.
 """
 
 from dataclasses import dataclass, field
+from math import isfinite
 from pathlib import Path
 from typing import Self
 
@@ -33,7 +40,11 @@ from monopriors.models.depth_completion.base_completion_depth import (
     PROMPT_INPUT_NAME,
     BaseCompletionPredictor,
 )
-from monopriors.models.zipdepth_checkpoint import load_zipdepth_state_dict
+from monopriors.models.zipdepth_checkpoint import (
+    load_zipdepth_checkpoint,
+    range_margin_from_checkpoint,
+    state_dict_from_checkpoint,
+)
 from monopriors.third_party.zipdepth.architecture import EncoderOutput, FeaturePyramid, ZipDepth, create_model
 from monopriors.third_party.zipdepth.model_utils import StateDict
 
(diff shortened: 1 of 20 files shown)


pablo-rerun:
**Status: trained, evaluated, and superseded — keeping as a draft.**

`zdpda-uw-v1` (this branch's lane: padded ultrawide prompt, 50/50 sampler, footprint-split eval) trained to
completion and was evaluated on the 20-segment ultrawide holdout at stride 10:

| checkpoint | wide AbsRel | wide edge | wide hard-20 | uw outside | uw inside | uw δ1 out |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| zdpda-v4 (wide only) | 0.0136 | 71 mm | 126 mm | 0.708 | 0.107 | 0.31 |
| zdpda-uw-v1 | 0.0173 | 88 mm | 155 mm | 0.234 | 0.096 | 0.69 |

It works: the ultrawide periphery goes from a near-uniform saturated field to a recognisable reconstruction of
the scene. The cost is a **27% wide regression** (0.0136 → 0.0173, with edge and hard-20 up in proportion), and
the gain plateaued — the 90k and 120k–130k checkpoints already sat at the final number. Inside the footprint
uw-v1 beats bilinear by only 5.6% (0.0960 against a 0.1017 prompt-upsample floor).

WiSE-FT interpolation between v4 and uw-v1 does not buy the wide lane back: **no alpha meets the wide gate**,
and the middle of the sweep is worse than either endpoint on both lanes at once.

**Superseded by the pre-fill direction.** Instead of training the model to cope with an unprompted periphery,
fill the periphery of the prompt: MoGe-2 metric depth on the same ultrawide RGB, scale-aligned to the wide LiDAR
over the footprint, with the real LiDAR block kept inside. On the **unchanged v4 weights** that scores
**0.141** outside (δ1 0.89) against uw-v1's fine-tuned 0.234 — and running uw-v1 with the same pre-filled prompt
scores 0.1438, i.e. *worse* than untouched v4. The ultrawide fine-tune adds nothing once the prompt is filled,
and it still costs the wide lane.

What is worth keeping from this stack is the lane and the evaluation, not the checkpoint: the footprint-split
metrics and the eval-only decomposition they enabled are what identified coverage as the binding constraint
(#216).

Full numbers, the seven-condition pre-fill table, the ceiling analysis and frame-for-frame pixel evidence:

The demo mode that produces those frames is #221.

pablo-rerun:
Closing as planned in the description: #215 and #216 now land on their own. #216 was rebased onto #215 (tree identical to this integration branch, proof: `git diff zipdepth/uw-3-run zipdepth/uw-2-lane` is empty), and #218/#219 were retargeted to `zipdepth/uw-2-lane`. Branch kept until the stack merges.

PR #219, opened 2026-09-04
monoprior/zipdepth: coverage-conditioned output-range margin

pablo-rerun:
Stacked on `zipdepth/uw-3-run`. Three commits: the coverage-conditioned output-range margin, a stricter ultrawide hole policy, and a correction to the run's flags after the gate exposed an inert filter pairing.

## Why (measured)

The prompted head emits `sigmoid(logits) * span + min`, with `min`/`max` taken from the prompt's own valid range. On an ultrawide sample the wide LiDAR prompt covers only the central 122x91 of the 192x256 canvas, so:

* On the 20 holdout segments a mean **27%** of periphery ground-truth pixels lie **outside** the prompt's `[min, max]` — 25% below the minimum, 3% above it — and **61%** of frames have more than 20% of their periphery unreachable.
* `zdpda-uw-v1` ran at `range_margin_m = 0.0` and therefore could not reach a quarter of the periphery at all.

The obvious fix does not work. A **global** absolute margin (PR #215's `range_margin_m`) gated at 3.9 m **broke the wide lane**: AbsRel **0.19** against **0.013** after 600 steps, because it re-spans the head for every frame including the fully-prompted wide ones.

## 1. Coverage-conditioned margin

`ZipDepthPrompt(range_margin_coverage_max=...)` makes the widening conditional on the prompt's own coverage of the canvas, computed inside the model from the validity mask it already builds:

```python
coverage_b = valid_prompt_mask.float().mean(dim=(1, 2, 3))
widen_b    = coverage_b <= self.range_margin_coverage_max
```

* The wide path is **bit-identical** to the un-widened head: `torch.where` hands back the very tensors the prompt-bounded head computes, and the span floor only ever binds on an excluded element, where it restores `span_b` exactly.
* The default `1.0` is the largest coverage any image can have, so it admits every image and reproduces the ungated margin exactly. `range_margin_m = 0.0` with `range_margin_coverage_max = 1.0` is bit-identical to v4, and the existing parity tests hold unchanged.
* The gate is a `torch.where` on per-element bounds, not Python control flow, so a mixed `--cameras both` batch resolves each element independently and the traced fp16 export graph keeps its data-independent shape. The trace-time `if self.range_margin_m > 0.0` branch structure is unchanged.
* Persisted exactly like `range_margin_m`: constructor keyword, checkpoint key, `resolved_config.json`, resume guard, init-checkpoint override notice, `load_zipdepth_prompt` override, and the ONNX cache filename (`..._m3.90_c0.60.onnx`). Checkpoints written before the gate read back as `1.0`, so they keep the behaviour they were trained with.

### The lanes separate cleanly, measured

Prompt coverage over the 4 gate segments (51 frames, stride 60) — the quantity the gate keys on:

| camera | mean | p05 | median | min | max | fraction ≤ 0.6 (widened) |
| --- | --- | --- | --- | --- | --- | --- |
| wide | 0.960 | **0.818** | 1.000 | 0.247 | 1.000 | **2.0%** (1 of 51) |
| ultrawide | 0.217 | 0.183 | 0.226 | 0.057 | **0.226** | **100%** |

The gap between the wide p05 and the ultrawide max is wide enough that any threshold in roughly `[0.23, 0.82]` behaves identically, so `0.6` is not a tuned constant and the gate is not sensitive to it.

### Gate results, 600 steps from `zdpda-v4`

4 holdout segments, stride 60, `--cameras both`:

| run | wide AbsRel | ultrawide outside-footprint AbsRel |
| --- | --- | --- |
| `zdpda-v4` reference | 0.012690 | — |
| uw-v1 control (margin 0.0) | 0.015466 | 0.4072 |
| global margin 3.9, ungated | ~0.19 | — |
| **uw-v2 gate** (3.9, coverage 0.6) | **0.024991** | **0.355270** |

The gated margin improves the periphery **12.7%** over the uw-v1 control and avoids the ungated margin's collapse of the wide lane. It does not hold the wide lane at v4's number; the residual 0.0123 gap is **not** mis-gating, since the coverage measurement shows only 2% of wide frames are widened at all. It is the cost of co-training the shared backbone on ultrawide frames with an opened head — `--cameras both` alone already accounts for 0.0028 (the uw-v1 control) — and 600 steps is a very early read on a 140k OneCycle schedule.

## 2. Ultrawide hole policy

`ultrawide_min_valid_fraction` counts invalid *area* only. It passes a frame where one window or skylight removes a contiguous fifth of the image while the rest is dense, and that is exactly the frame the metric loss cannot supervise over a whole region.

`ultrawide_max_hole_fraction` drops a frame whose **largest connected** invalid region, measured before erosion like the valid fraction, exceeds the given fraction of the frame. Over 5,415 ultrawide frames the two filters reject overlapping but different sets: **6.6%** fall below 0.8 valid, **3.5%** carry a hole larger than 20% of the frame.

Components come from `scipy.ndimage.label` with 4-connectivity on the CPU; a CUDA-built mask is copied down first, so both builders decide identically through the one shared policy function. The labelling pass is gated behind the total invalid area, which no single region can exceed, so the default `1.0` never labels and costs nothing. Rejections are counted as `skipped_large_hole_frames` and logged beside `skipped_low_valid`.

### The cap is inert at or above `1 - min_valid_fraction`

A frame that clears the valid-fraction gate has at most `1 - min_valid` invalid area *in total*, and no single region can exceed the total, so a larger cap can never fire. The first uw-v2 gate ran `0.8`/`0.2` and logged `skipped_large_hole = 0` across all 600 steps. `UltrawidePolicy.__post_init__` now **refuses** such a pairing rather than silently doing nothing; both ways of meaning "no hole filter" (the `1.0` default and an explicit `1.0`) still stand.

The run therefore uses **`--ultrawide-min-valid-fraction 0.7 --ultrawide-max-hole-fraction 0.2`**, which also keeps uw-v1's valid threshold so the margin's effect stays attributable. Re-gated at 300 steps it rejected **17** frames for one oversized hole beside 39 for low valid fraction — the ~3.5% tail the filter exists for — and wide AbsRel came in at 0.027725, well clear of the 0.05 break threshold.

## Tests

New behaviour tests cover: the two lanes' coverages against the 0.6 threshold; bit-identity of a full-coverage prompt under margin 3.9 + gate 0.6; a padded 122x91 prompt opening the full `[0.1, 4.0] m` window; a mixed batch resolving both elements independently; the default 1.0 still widening every image; constructor validation; checkpoint round-trip and the pre-gate default; resume refusal on a changed gate; one ONNX graph per (margin, gate) pair; `largest_hole_fraction` separating speckle from one block at equal invalid area; the builder rejection and its counter; the default hole policy keeping a half-hole frame; and refusal of a subsumed hole cap.

Verified on the dl-server:

| env | lint | typecheck | tests | deadcode |
| --- | --- | --- | --- | --- |
| `zipdepth-dev` | pass | 0 errors | 114 passed, 5 skipped (base: 103 passed, 5 skipped) | clean |
| `zipdepth-catalog-dev` | pass | — | 161 passed, 2 skipped | — |
| `monoprior-dev` | pass | 0 errors | 130 passed, 18 deselected | clean |

The catalog-lane tests (`test_catalog_ultrawide.py`, including the hole-policy and subsumption tests) `importorskip` outside `zipdepth-catalog-dev`, which is why the new hole tests only move that env's count.

`tests/test_infer_rerun_pixels.py` is excluded in the zipdepth envs: it imports `omegaconf` through `monopriors.rr_logging_utils`, which those envs do not carry. That collection error is pre-existing on the base branch.


pablo-rerun:
**Verdict: do not merge as-is.** `zdpda-uw-v2` (this branch's mechanism, `range_margin_m = 3.9`,
`range_margin_coverage_max = 0.6`) was trained and evaluated on the 20-segment ultrawide holdout. It regressed
on **every** axis against uw-v1:

| checkpoint | wide AbsRel | wide edge | wide hard-20 | uw outside | uw inside |
| --- | ---: | ---: | ---: | ---: | ---: |
| zdpda-v4 | 0.0136 | 71 mm | 126 mm | 0.708 | 0.107 |
| zdpda-uw-v1 | 0.0173 | 88 mm | 155 mm | 0.234 | 0.096 |
| zdpda-uw-v2 | 0.0223 | 91 mm | 156 mm | 0.287 | 0.243 |

Ultrawide **inside**-footprint AbsRel 0.096 → 0.243 is the worst of it: that is now worse than the 0.1017
bilinear prompt-upsample floor, i.e. the model under-uses a prompt it is sitting directly on. The frames show
the matching signature — an over-smoothed periphery where uw-v1 had distinct surfaces.

**The gate misfires, and the misfires are the entire wide loss.** The coverage gate fired on **26 wide frames**
(1.8% of the wide holdout) where the LiDAR prompt happened to be sparse enough to look ultrawide. Excluding
those frames the wide lane scores **0.0167** — no worse than uw-v1. So the wide regression is not a capacity
trade, it is a gate that cannot tell the two lanes apart from prompt coverage alone.

**Two independent lines of evidence argue against range widening as the fix at all:**

- *Decomposition.* Under a full-coverage oracle prompt, with the same output-range cap in force, v4 scores
  0.050 and uw-v1 0.040 outside. The cap is real (28.7% of periphery pixels are outside the prompt's observed
  range) but it is not what holds the padded number at 0.708 — missing coverage is.
- *Literature.* Every partial-coverage method that handles this (DepthLab, OMNI-DC, Depth Prompting, Any2Full)
  either randomizes the normalization window during training or removes the bound entirely. None uses a
  coverage-dependent switch. A runtime statistic selecting between two decoder contracts is exactly the design
  this PR introduces, and uw-v2 is what it costs.

**What replaced it.** Pre-filling the ultrawide prompt with MoGe-2 metric depth, scale-aligned to the wide LiDAR
over the footprint, takes v4's outside AbsRel from 0.708 to **0.141** with **no training and no range change**,
against uw-v1's fine-tuned 0.234. Details, tables and pixel evidence:

**Checkpoint-version footgun worth fixing before anything here lands.** uw-v2's checkpoint records
`range_margin_coverage_max`, but the `ZipDepthPrompt` wrapper on the sibling demo branch predates that field —
loading uw-v2 through it widens the range for *every* image, including the fully prompted wide lane. Measured on
identical frames: wide AbsRel 0.0162 → 0.2712 (42446050) and 0.0303 → 0.3401 (44796684), purely from the wrapper
mismatch. A checkpoint carrying a field the loading code does not understand must fail loudly rather than
silently change behaviour.

If the branch is kept, the recommendation is to keep #215's plain `range_margin_m` mechanism (default off) and
drop the coverage conditioning; if the range bound is revisited later, the randomized-normalization-window
route is the one with support behind it.

PR #224, opened 2026-09-06
dataforge: msd ground-truth layer, measured world axes, three-headset run (msd 3/3)

pablo-rerun:
Third of three. Stacks on #223. Adds the `gt` layer for the Monado SLAM Datasets and finishes the end-to-end path: convert → two rrds per sequence on the NAS → register both layers → view.

## What changes

- **`gt` layer** written by the same `convert` call, same recording id, at `<output_root>/gt/`: temporal `world_T_rig` on `/world/rig_00` at the full GT rate (quaternion reordered from the file's wxyz), the full path as a static `LineStrips3D` at `/world/runs/gt/trajectory`, a cursor-relative trail as columnar `Points3D` at `/world/runs/gt/trail`, and the root `ViewCoordinates`, which this layer owns. Degenerate quaternions are sanitized to identity and counted into the `gt` recording properties.
- **World axes measured, then fixed per device.** `measured_world_up` rotates the first two seconds of accelerometer samples by the nearest GT rotation and reports the dominant world axis. All three headsets measure **+y** (Index 0.96 g, G2 0.98 g, Odyssey+ 0.93 g, off-axis ≤ 0.1 m/s²), so every device declares `world_up="+y"` → `RIGHT_HAND_Y_UP`; a sequence that disagrees warns instead of reorienting itself.
- `convert` skips only when both layers exist and rebuilds both otherwise; `register` output counts both layers.

## Evidence

- 124 tests pass, lint/typecheck/deadcode clean, lock untouched.
- Real runs onto `/mnt/nas/datasets/msd-rrd`: Index MIO09 (2×186 samples, 3182 poses), G2 MGO09 (4×107, 179 mag rows, 3264 poses), Odyssey+ MOO09 (2×147, 247 mag rows, 4545 poses); no raw left on disk.
- Registered on the local catalog (`msd-index`, `msd-g2`, `msd-odyssey`, base + gt layers, default and table blueprints).
- Pixel check in the AV1-capable viewer with base + gt loaded together: the rig rides the GT pose, the trajectory line and trail render, both cameras decode; frames at 1.5 s and 3.0 s differ as expected for the up-down motion (`/tmp/rerun-viewer-validation/msd-gt-layer/`).

## Known gaps

- The Index GT starts 8.5 ms after the earliest sensor sample, so the rig has no pose for that first window (children sit at the origin until the first pose). No resampling by design.
- GT runs at ~915–925 Hz rather than the nominal 1 kHz on all three devices (missing rows, not degenerate ones).
- The default 3D view's initial eye frames the scene tightly; fine for scrubbing, worth a hand-set eye later.


pablo-rerun:
Follow-up commit d4c8a02: the Follow view's eye was RoboCap's IMU-frame constant, which put the chase camera beside these headsets with a rolled horizon. It is now derived per device from the front stereo pair's calibration: forward = mean optical axis, up = baseline × forward (image-up is 90° off on the G2, whose cameras are mounted rolled). Constants live on `MsdDevice.follow`; convert warns if a sequence's calibration disagrees by more than 5°. Pixel-checked on all three headsets after re-converting one short sequence each (`/tmp/rerun-viewer-validation/msd-follow/`): eye behind and above, level horizon, trail along the rig's up. Catalog defaults re-registered; the full corpus is being re-converted so the embedded blueprints match.

pablo-rerun:
fb4ec83: camera nodes now carry `camera_model` (`kb4` / `pinhole-radtan8`) and, for radtan8 cameras, `distortion_valid_radius` (basalt's `rpmax`) as optional AnyValues, so nothing from the upstream calibration is lost. Decision recorded here: each device keeps its upstream default `calibration.json` (kb4 on the Index, radtan8 on the G2 and Odyssey+, the model their factory calibration and basalt's published results use); the alternative `calibration.kb4.json` files are not used. G2 and Odyssey+ are being re-converted so the stored rrds carry the new keys.

pablo-rerun:
Simplify pass (six commits, fb4ec83..9e0e07fd) from a four-reviewer cleanup (reuse / quality / efficiency / altitude), behaviour-preserving:

- Shared writers in `logging_toolkit`: `log_camera_node`, `log_pose_track`, `time_column` (zero-copy index), private `_log_sensor_node` / `_log_scalar_channel` used by `log_imu` and `log_magnetometer`; `require_av1_nvenc` cached; the `log_video_stream` tap no longer pays the record-batch conversion on the pre-existing shift path and retimes the keyframe chunk with `searchsorted` instead of a per-sample dict.
- `dataforge/blueprints.py`: one rig/table blueprint skeleton and eye factory for robocap and msd (blueprint object trees verified identical before and after for all 14 blueprints).
- `msd.py` 1406 → 857 lines: `dataforge/archives.py` (zip / 7-Zip volume readers), `dataforge/basalt.py` (calibration json → simplecv cameras, follow frame), `dataforge/euroc.py` (csv readers, gt trajectory, measured world up); `convert` is `read_sequence` → `write_base_layer` → `write_gt_layer` over typed intermediates; raw-budget tally is one scandir pass that excludes HF's `.cache` and always checks leftovers; HF revision resolved before the recording opens.
- `paths.LAYERS` drives register and view (view now loads every layer); `schema.GT_RUN_SOURCE`; `writing.atomic_recording` defaults the application id.
- Deps: `ffmpeg >= 9.0.1` declared for dataforge (the env's ffmpeg has `av1_nvenc`, so `DATAFORGE_FFMPEG` is a genuine override); `RERUN_INSECURE_SKIP_HOST_CHECK` on the catalog tasks; beartype `claw_is_pep526=False` for the array-heavy loops.
- Tests: `tests/conftest.py` holds the shared NVENC fixture, read-back helpers and frame generator; a module-scoped converted fixture replaces eight redundant GPU encodes; four tautological assertions removed. 144 tests, all gates green.

Verified end to end: re-converting MIO09 with the cleaned code reproduces the stored recording's entities, row counts and timestamps for both layers. Held for discussion (RISKY): decimating the static gt trajectory strip, prefetching the next archive during encode, parallel per-camera encodes, keeping clips on failure, centralizing the two-target skip rule in the base class.

pablo-rerun:
Thermo-nuclear review (Codex, gpt-5.6-sol, read-only over the full stack diff) returned **request changes** with nine findings. Applied in eight commits, 9e0e07fd..61538767, 177 tests, all gates green, rrd identity proven before/after each phase:

- **Validated camera record** (`basalt.CalibratedCamera`, `Kb4Intrinsics` / `Radtan8Intrinsics`, `load_calibration`) replaces three index-aligned lists plus MSD's parallel clip/timestamp tuples; missing coefficients, wrong lengths or camera counts now fail loudly instead of becoming zeros. The three real device calibrations are checked in as fixtures, which also un-circularizes the Follow-frame test.
- **Retiming tap classifies chunks explicitly** (codec / sample / keyframe) and raises on anything else; the remux tests run on a checked-in 48-frame AV1 clip, so they no longer need a GPU.
- **`dataforge/video_encoding.py`** holds ffmpeg/NVENC; `logging_toolkit` re-exports the contract and keeps the Rerun writers.
- **One HF commit per conversion**: the sha is resolved once and used for listing, fetches and provenance.
- **Layer rule** (README "### The layer rule", cherry-picked from the Lamaria stack): MSD keeps `gt.csv` as a sidecar under `<root>/sidecars/<recording_id>/`, the gt layer rebuilds from the sidecar plus the accelerometer read back from the base rrd, skips are per layer, base and gt are staged and published back to back, gt is written with `send_properties=False` and carries no RecordingInfo, and `register --replace` exists for the in-memory server.
- **Types**: `CameraKind` / `PeerSensorKind` in simplecv with `SensorKind` as their union; `log_camera_node` takes explicit `camera_model` and `distortion_valid_radius`. A non-positive radtan8 `rpmax` (the Odyssey+ ships 0.0) now means "no stated limit" and the key is omitted.
- **Decomposition**: `msd.py` 658 lines + `msd_layers.py` 569; tests split the same way. Docs trimmed to one home each (`docs/msd.md` for the measurement evidence).

Declined, with reasons: a declared-layers base class with a completion manifest (rejected earlier as overcomplicating; the crash window is covered by staged publication instead), and renaming `ImuChannel` / turning `FrameSource` into a tagged union (both are the Lamaria contract).

The stored corpus is being regenerated so every sequence carries the sidecar, the RecordingInfo-free gt layer and the camera-model keys.

pablo-rerun:
Two more changes after the review passes, a30c45fd..10ad241c (219 tests green):

- **Frames are encoded upright.** Each camera is rotated by the quarter turn that aligns its image-up with the headset's up, derived from the calibration (`basalt.upright_quarter_turns`), inside ffmpeg's `transpose` filter (`encode_frames_to_mp4(rotate_cw_quarter_turns=)`), and its calibration is rotated with the pixels (`basalt.rotate_camera_cw`: focal lengths and principal point swapped, camera frame rolled about the optical axis, radtan8 tangential terms remapped as p1' = p2, p2' = −p1). Verified numerically against `cv2.projectPoints` for both models: 1e-13 px. The Reverb G2 is the only device affected: its front pair rolls 90° and its side pair 270°, which is why the rule is per camera; the Index and Odyssey+ derive 0 and are byte-identical. The turn is recorded as `image_rotation_cw_deg` on the camera node.
- **Trail as line segments.** The Follow view's trail was 2 cm dots that read as scattered balls. It is now one two-point `LineStrips3D` per pose (`logging_toolkit.log_trail_segments`, shared with Lamaria) with a ui-point radius, so the cursor-relative window joins into a continuous tail at any pose rate; the Follow view shows the full trajectory thin and dim for context instead of hiding it.

The layer rule paid off on the first try: rebuilding a gt layer from its sidecar and base rrd took 3 s with no archive fetch. The G2 corpus is being re-converted for the upright frames.

PR #245, opened 2026-09-10
slam-rs: select corners per grid cell on the GPU

pablo-rerun:
Detection on the GPU lane downloaded the candidate image and its column bitmask —
0.92 MB + 0.115 MB per camera, **2.07 MB per two-camera frameset** on MIO10 — and
then did all of the selection on the host: a row band per cell row and rung, a
column filter, OpenCV's non-maximum suppression per cell, a sort and three gates.
The three FAST kernels were 0.04 ms of a 1.45 ms stage.

This moves the selection to the device. One cube per grid cell picks that cell's
winner and the readback is one packed key per cell — 361 x 4 B per camera instead
of 1,036,800.

### Why one key is the whole answer

The threshold ladder cannot change which corner it is. A candidate at rung `t` is
`kept > t`; `suppress_non_maxima` kills a pixel only through an in-window
neighbour scoring at least as much, and such a neighbour is itself a candidate at
every rung the pixel is. Suppression therefore does not depend on the rung, and
the ladder only admits survivors in descending score. With
`optical_flow_detection_num_points_cell = 1`, which every shipped config sets,
the cell's outcome is the best survivor over the floor that clears `safe_radius`,
the masks and `EDGE_THRESHOLD`.

The packed key `((255 - score) << 24) | (y << 12) | x` reduces under integer
**minimum** to exactly the host's total order — score descending, then row, then
column, which is what the row-major band walk and a stable sort produce. Integer
minimum is associative, commutative and exact, so the shared tree agrees with the
host bit for bit and a subgroup fast path would agree with both.

The masks stay on the host and are exact there: `cam0OverlapCellsMasksForCam`
pushes `cell` x `cell` rectangles at the cell origins and a cell's candidates lie
strictly inside it, so a cell is wholly masked or wholly clear. The guard is at
the camera level — a rectangle that does not name one cell of *this* camera's
grid sends the whole camera down the band walk, which is what the mixed-geometry
rigs the port supports deliberately produce. So do `num_points_cell != 1`, a cell
no wider than the FAST ring, and a frame 4,096 pixels or more on a side.
`CpuCornerScan` takes the trait's default and is untouched.

### Harness

`/tmp/slam-rs-s32/ab/ab.sh`, one core, interleaved rounds against
`slam-rs/stage-timers` (44cbdb0f). Core measured: `809e978b`, whose tree differs
from this branch's head only by test-file formatting and `docs/design-notes.md` —
neither is compiled into the core.

**MIO10, 3 rounds**

| Core | median ms | mean ms | p95 ms | max ms | ATE vs GT cm | ATE vs C++ cm | tracked/lost | byte-identical | state-identical |
|---|---:|---:|---:|---:|---:|---:|---:|---|---|
| base | 5.151 | 5.688 | 9.670 | 11.236 | 1.504 | 0.311 | 412/0 | Y | Y |
| this | 4.372 | 4.835 | 8.744 | 10.349 | 1.504 | 0.311 | 412/0 | Y | Y |

| Core | pyramid | **detect** | track | stereo | optimize | linearize | solver | marginalize | measure | track_ms | glue |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| base | 0.173 | **1.466** | 0.756 | 0.212 | 2.243 | 0.746 | 1.196 | 0.148 | 2.465 | 5.151 | 0.152 |
| this | 0.172 | **0.589** | 0.741 | 0.188 | 2.269 | 0.753 | 1.172 | 0.148 | 2.462 | 4.372 | 0.152 |

`ACCURACY PASS; IDENTICAL yes; delta -0.778 ms (-15.11%) vs base.`
Gate: ATE 1.504 cm against a 1.654 cm band, zero lost framesets, median gain
0.778 ms against a 0.100 ms bar. **PASS.**

**MGO09, four 640x480 cameras, 1 round**

| Core | median ms | ATE vs GT cm | tracked/lost | byte-identical | state-identical |
|---|---:|---:|---:|---|---|
| base | 16.904 | 0.770 | 107/0 | Y | Y |
| this | 13.165 | 0.770 | 107/0 | Y | Y |

`frontend_detect` there is 2.233 against 2.241 — **unchanged**. On a small
four-camera rig the readback was never the cost; four device round trips is what
2.24 ms is. The `track_ms` and estimator movement in that single 107-frameset
round is run noise, not this change: the trajectories are byte-identical.

A hoisted variant that launched every camera and read them all in one download
was written, measured and dropped — right in isolation (0.502 ms against 0.822
for two 960x960 cameras through the scanner; 0.74-0.84 against 1.01-1.04 per
frameset through the whole frontend in one process) and wrong under the gate,
twice (`frontend_detect` 1.418 and 1.422 ms, `track_ms` -8.4% and -8.8%). A
counter through the real frontend rules out the obvious cause — the batch
answered every selection. D72 records it so it is not rebuilt blind.

### Tests

`tests/gpu_detect.rs` is the exactness gate rather than the A/B run: the whole of
`detectKeypointsWithCells` runs twice over the same frame — through the band walk
and through the device selection — and the two `KeypointsData` must be equal
corner for corner and response for response, on the committed 960x960 MIO10
frames, both cameras, empty and half-occupied, at three safe radii, under
cell-aligned masks and one that straddles a boundary, at three budgets, and on
three clamped grids where the last cell runs past the image. It is a second test
binary because `the_whole_gpu_path_holds_the_pool_flat` asserts an exactly flat
CubeCL pool and every test in one binary shares one client; two 960x960 scanners
beside it failed that about one run in three.

`cargo test --workspace` green; `slam-rs-wgpu-test` 20 + 3 green; `pytest` 256
passed, 1 skipped; both Clippy lanes clean at `-D warnings`.

No `.rrd` is produced by this change and none is needed.

### DIFF
--- a/packages/slam-rs/crates/slam-rs/src/frontend/detect.rs
+++ b/packages/slam-rs/crates/slam-rs/src/frontend/detect.rs
@@ -55,7 +55,11 @@
 //! parity gate seeds the tracker with the C++ keypoints for that reason.
 
 use kornia_image::{Image, ImageSize};
-use kornia_imgproc::features::{FastCorner, Rect as KorniaRect, fast_detect_rect_u8};
+use kornia_imgproc::features::{Rect as KorniaRect, fast_detect_rect_u8};
+
+/// kornia's FAST corner, re-exported because [`CornerScan::band`] hands it back:
+/// a backend outside this crate cannot implement the trait without naming it.
+pub use kornia_imgproc::features::FastCorner;
 
 use crate::image::ImageU16;
 
@@ -366,6 +370,27 @@ pub struct DetectorConfig {
 /// 1 and no further, rather than a wedged process (decision D32).
 pub const LOWEST_THRESHOLD_RUNG: i32 = 1;
 
+/// The thresholds [`detect_keypoints_with_cells`]'s ladder visits, in order.
+///
+/// `max_threshold`, then halved by integer division for as long as the value
+/// stays at or above the floor, which is `min_threshold` but never under
+/// [`LOWEST_THRESHOLD_RUNG`]. The shipped 40/5 configs give 40, 20, 10, 5; 40/6
+/// gives 40, 20, 10 and **not** 6, because the next halving is 5 and the ladder
+/// never visits the floor itself unless a halving happens to land on it. The
+/// sequence is empty when `max_threshold` is already under the floor.
+///
+/// One iterator rather than two expressions because both callers need it and
+/// they must not drift: the cell walk steps through it, and
+/// [`CornerScan::select_cells`] is handed its **last** value, which is the only
+/// rung that decides a one-point-per-cell winner. A device handed
+/// `min_threshold` instead would admit corners scoring between the two, which
+/// the walk never sees.
+pub fn threshold_rungs(config: &DetectorConfig) -> impl Iterator<Item = i32> {
+    let floor: i32 = config.min_threshold.max(LOWEST_THRESHOLD_RUNG);
+    std::iter::successors(Some(config.max_threshold), |threshold| Some(threshold / 2))
+        .take_while(move |threshold| *threshold >= floor)
+}
+
 /// One cell row's raw FAST candidates at one threshold, over the whole image width.
 ///
 /// `fast_detect_rect_u8` detects over **whole rows** and filters the result by
@@ -464,8 +489,9 @@ pub struct BandRequest {
 pub struct CellSelect {
     /// The detected image's own grid, as [`detect_keypoints_with_cells`] derives it.
     pub grid: CellGrid,
-    /// The rung the ladder stops at, which is the only one that decides the
-    /// winner (see [`CornerScan::select_cells`]).
+    /// The **last rung the ladder visits** ([`threshold_rungs`]), which is the
+    /// only one that decides the winner (see [`CornerScan::select_cells`]).
+    /// Not `min_threshold`: the halving ladder need never reach it.
     pub threshold: i32,
     /// `optical_flow_image_safe_radius`; `0` switches the gate off.
     pub safe_radius: f32,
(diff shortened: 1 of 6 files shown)


pablo-rerun:
Fix round for the two review findings, on `0e2c7f20` and `f751ff66`.

**P1 — the device got the configured lower bound, not the last rung the host visits.**
`threshold_rungs(config)` is now the one place the ladder is computed: the cell
walk steps through it and the device path is handed its `.last()`, so the two
cannot drift. With 40/6 the walk visits 40, 20, 10 and the device now stops at
10; before this it admitted anything over 6, which on a committed MIO10 frame is
62 corners against the walk's 54. A ladder with no rung at all — a maximum
already under the floor — returns the empty answer the walk would give without
touching the frame. `CellSelect`/`select_cells`, the kernel's geometry field and
D72 no longer call that bound a floor.

**P2 — the "clamped grids" test never clamped.** `CellGrid::new` floors and
centres, so none of its grids has a cell past the image and neither strict clamp
in `fast_cell_select_kernel` ran. The three frames are kept for what they do
test (kornia's in-block filter, an uneven grid start) and renamed to say so, and
two caller-supplied grids now reach the clamps: a last column and row that
overhang the frame, and a last column whose candidate window is empty and has to
come back as the no-winner sentinel. Both preconditions are asserted in the test.
The clamped columns lie past `width - EDGE_THRESHOLD - 1`, so no corner can come
out of them on either lane — what the equality proves there is that the device
stays inside the image and sees the same zero rim.

**Also**, from the review's "not covered" list: the frame at `CELL_KEY_LIMIT`,
`num_points_cell = 2` and four camera slots through one reused scanner. The GPU
scanner is now wrapped in a counting one, because equality alone cannot tell the
two paths apart — a device path that quietly never engaged agrees with the host
walk perfectly — so every case asserts which path it meant. That turns the three
fallbacks (straddling mask, two points per cell, key limit) into assertions.

`cargo test --workspace` green (25 binaries); `gpu_kernels` 20 + `gpu_detect` 8
passed; `cargo clippy --workspace --all-targets -- -D warnings` clean on both
lanes.

MIO10, 3 interleaved rounds, one core, base = main + timers (44cbdb0f):

| Core | median ms | mean ms | p95 ms | ATE vs GT cm | tracked/lost | frontend_detect ms | byte-identical |
|---|---:|---:|---:|---:|---:|---:|---|
| base | 5.110 | 5.632 | 9.688 | 1.504 | 412/0 | 1.423 | Y |
| detect-gpu-fix | 4.371 | 4.831 | 8.761 | 1.504 | 412/0 | 0.576 | Y |

`ACCURACY PASS; IDENTICAL yes; delta -0.739 ms (-14.47%) vs base` — unchanged, as
expected: msdmi is 40/5, whose ladder lands on its own minimum, so the fix cannot
move MIO10.

PR #247, opened 2026-09-10
slam-rs: redetect on demand in the fast profile

pablo-rerun:
## What

`add_points` ran on every frameset (basalt's schedule). cuVSLAM detects only
once its survivors fall under a fraction of what the last detection left it, and
so pays detection about every fourth frame. This adds that rule as
`port.redetect_survivor_ratio` and turns it on in the fast profile alone.

- `crates/slam-rs/src/frontend/flow.rs` — `should_detect()` gates `run_passes`'
  call to `add_points`; `last_detect_count` is camera 0's count as the last
  detecting frameset left it, snapshotted in `FrameState` so a refused frameset
  does not move it.
- `crates/slam-rs/src/config.rs` — the `port.redetect_survivor_ratio` field,
  `0.0` (= basalt: detect every frameset) by default, skipped on serialization
  while it is off.
- `slam_rs/reference.py` — `PORT_CONFIG_KEYS`; the overlay may insert a
  port-only key that no vendored config carries, and any other unknown key is
  still a `KeyError`.
- `configs/profiles/fast.json` — `"port.redetect_survivor_ratio": 0.7`.
- `docs/design-notes.md` — D75 and its index line.

**No vendored config changed.** basalt has no field for this knob, so it is
spelled `port.` rather than `config.` and the vendored `configs/*.json` stay the
documents the C++ reference runs read, key for key —
`test_the_vendored_configs_are_the_ones_the_cpp_runs_used` is untouched and
green.

**The gate is one decision for the whole rig, taken on camera 0.** `add_points`
is camera 0's detection, the cross-camera match that carries its new ids into
cameras 1..n, and the non-overlap pass on those cameras, so gating part of it
would leave a rig half detected. MGO09 (4 cameras) exercises that path.

## MIO10, 3 rounds, fast profile, `--base wave1`

| Core | profile | median ms | mean ms | p95 ms | max ms | ATE vs GT cm | ATE vs C++ cm | tracked/lost | byte-identical to base | state-identical | cores used |
|---|---|---:|---:|---:|---:|---:|---:|---:|---|---|---:|
| base (wave1) | fast | 3.526 | 3.885 | 6.039 | 7.396 | 1.525 | 0.311 | 412/0 | Y | Y | 0.877 |
| redetect | fast | 2.974 | 3.172 | 4.988 | 5.894 | **1.447** | 0.651 | 412/0 | N | N | 0.906 |

| Core | pyramid | detect | track | stereo | predict | keyframe | optimize | linearize | solver | marginalize | measure | track_ms | glue |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| base (wave1) | 0.143 | 0.574 | 0.751 | 0.193 | 0.006 | 0.000 | 1.514 | 0.535 | 0.753 | 0.106 | 1.683 | 3.526 | 0.150 |
| redetect | 0.144 | **0.000** | 0.755 | **0.001** | 0.006 | 0.000 | 1.394 | 0.479 | 0.721 | 0.097 | 1.519 | 2.974 | 0.133 |

Round medians: base 3.526, 3.562, 3.581; redetect 2.989, 2.974, 3.023.
Delta **-0.551 ms (-15.63 %)**. The median frame now detects nothing, so both
`detect` and the `add_points` half of `stereo` fall off it.

## MGO09, 1 round, fast profile, `--base wave1`

| Core | profile | median ms | mean ms | p95 ms | max ms | ATE vs GT cm | ATE vs C++ cm | tracked/lost |
|---|---|---:|---:|---:|---:|---:|---:|---:|
| base (wave1) | fast | 10.357 | 10.083 | 12.410 | 13.043 | 0.766 | 0.317 | 107/0 |
| redetect | fast | 8.536 | 8.280 | 11.054 | 11.710 | **0.757** | 0.317 | 107/0 |

Delta -1.821 ms (-17.58 %).

## Keyframe cadence, landmarks, tracked keypoints (post-warmup, round 1)

| clip / core | frames | keyframes | frames per keyframe | detect every N framesets | mean landmarks | mean tracked kpts | mean observations |
|---|---:|---:|---:|---:|---:|---:|---:|
| MIO10 base (wave1) | 352 | 50 | 7.04 | 1.00 | 46.0 | 119.1 | 321.2 |
| MIO10 redetect | 352 | 48 | **7.33** | 2.91 | 38.5 | 84.3 | 288.2 |
| MGO09 base (wave1) | 47 | 7 | 6.71 | 1.00 | 106.2 | 306.3 | 879.6 |
| MGO09 redetect | 47 | 7 | **6.71** | 5.22 | 101.9 | 231.6 | 848.9 |

The cadence stays inside the 5-9 band on both clips, so the vote the later
scheduling work is priced against has not moved.

## Gate

| clause | limit | MIO10 | verdict |
|---|---|---|---|
| ATE vs GT | <= 1.654 cm | 1.447 cm | PASS |
| lost framesets | 0 | 0 (412 tracked) | PASS |
| median gain vs wave1 | >= 0.1 ms | 0.551 ms | PASS |
| run-to-run identical within the label | required | trajectory md5 and state digest identical over 3 rounds | PASS |
| MGO09 band | <= 0.8466 cm | 0.757 cm | PASS |

Trajectories are **not** byte-identical to the stack, as expected for a
trajectory-changing lever.

## Why 0.7 and not 0.5

`0.5` was measured first and also clears the gate — MIO10 median 2.337 ms
(-1.212), ATE 1.516 cm; MGO09 0.777 cm — but it detects only every 6.6
framesets on MIO10 and every 23.5 on MGO09, taking mean landmarks to 25.7 and
tracked keypoints to 58.4 on MIO10 and MGO09's keyframe cadence to 9.40. `0.7`
gives 0.582 ms against the 0.3 ms this lever was asked for and is the only
setting where **both** measured clips score better than the stack. D75 records
the `0.5` numbers so the trade is re-openable once the wider clip set has run.

## Tests

`cargo test --workspace --release`: 447 passed, 0 failed.
`cargo clippy --workspace --all-targets -- -D warnings`: clean.
`ruff check .`: clean.
Python suite: 261 passed, 1 skipped, 18 deselected (with `PYTHONPATH` set to
this worktree — the shared bench env's editable `.pth` otherwise wins).
`cargo fmt --all --check` reports only the pre-existing `estimator/mod.rs:1287`
diff, left alone.

New tests: `the_default_config_detects_on_every_frameset`,
`a_survivor_ratio_nothing_reaches_detects_only_on_the_first_frameset`,
`a_skipped_frameset_leaves_no_camera_detected`,
`the_redetect_knob_is_off_and_unwritten_in_every_shipped_config`,
`the_redetect_knob_survives_the_round_trip_when_a_profile_sets_it`,
`test_no_vendored_config_carries_a_port_key`,
`test_an_invented_port_key_is_rejected`.

## Caveats

- Only MIO10 and MGO09 are measured. MIO11, MIO07, MGO07, MGO10 and MGO11 are
  unrun; the clips with fast-dying tracks are where a lower steady-state
  landmark count would show.
- MIO10's ATE vs the C++ trajectory moves 0.311 -> 0.651 cm. Still well inside
  the 2 cm path bound, but it is a real move away from basalt's own path.
- No `.rrd` was produced by this work.

### DIFF
--- a/packages/slam-rs/configs/profiles/fast.json
+++ b/packages/slam-rs/configs/profiles/fast.json
@@ -1,1 +1,1 @@
-{"config.vio_max_iterations": 4}
+{"config.vio_max_iterations": 4, "port.redetect_survivor_ratio": 0.5}
--- a/packages/slam-rs/configs/profiles/fast.json
+++ b/packages/slam-rs/configs/profiles/fast.json
@@ -1,1 +1,1 @@
-{"config.vio_max_iterations": 4, "port.redetect_survivor_ratio": 0.5}
+{"config.vio_max_iterations": 4, "port.redetect_survivor_ratio": 0.7}
--- a/packages/slam-rs/crates/slam-rs/src/config.rs
+++ b/packages/slam-rs/crates/slam-rs/src/config.rs
@@ -581,10 +581,10 @@ mod tests {
         let base: VioConfig = VioConfig::from_json_str(MSDMI_JSON).unwrap();
         let overlaid: String = MSDMI_JSON.replace(
             "\"value0\": {",
-            "\"value0\": {\"port.redetect_survivor_ratio\": 0.5,",
+            "\"value0\": {\"port.redetect_survivor_ratio\": 0.7,",
         );
         let config: VioConfig = VioConfig::from_json_str(&overlaid).unwrap();
-        assert_eq!(config.port_redetect_survivor_ratio, 0.5);
+        assert_eq!(config.port_redetect_survivor_ratio, 0.7);
 
         let mut normalised: VioConfig = config.clone();
         normalised.port_redetect_survivor_ratio = 0.0;
--- a/packages/slam-rs/crates/slam-rs/src/config.rs
+++ b/packages/slam-rs/crates/slam-rs/src/config.rs
@@ -100,6 +100,16 @@ pub enum ConfigError {
     Parse(#[from] serde_json::Error),
 }
 
+/// Whether [`VioConfig::port_redetect_survivor_ratio`] is at its off value.
+///
+/// The port's own key is skipped when it is off, so a basalt document still
+/// round-trips to exactly the keys it arrived with and a C++ run reading a
+/// config the port wrote back never meets a key cereal has no field for.
+#[expect(clippy::trivially_copy_pass_by_ref, reason = "serde's predicate shape")]
+fn redetect_is_off(ratio: &f32) -> bool {
+    *ratio == 0.0
+}
+
 /// cereal's outer wrapper: every basalt JSON is one object under `value0`.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 struct Value0<T> {
@@ -181,6 +191,26 @@ pub struct VioConfig {
     /// Per-level residual cap for accepting a recall.
     #[serde(rename = "config.optical_flow_recall_max_patch_norms")]
     pub optical_flow_recall_max_patch_norms: Vec<f32>,
+
+    // ── the port's own knobs ────────────────────────────────────────────
+    /// Survivor fraction below which a frameset detects; `0` detects always (D75).
+    ///
+    /// The one key here that basalt has no counterpart for, which is why it is
+    /// spelled `port.` instead of `config.`: the vendored `configs/*.json` are
+    /// the files the C++ reference runs read, so nothing writes a key into them
+    /// that the C++ never saw, and this one is carried by a profile overlay
+    /// alone (`configs/profiles/fast.json`). `0.0` — the value every basalt file
+    /// leaves it at, and [`VioConfig::default`]'s — is basalt's own behaviour:
+    /// `addPoints` on every frameset. Above zero the frameset detects only once
+    /// camera 0 holds fewer than this fraction of the keypoints the last
+    /// detecting frameset left it with, which is cuVSLAM's rule. Anything not
+    /// finite and above zero reads as `0`, so a garbled value detects rather
+    /// than silently stopping.
+    #[serde(
+        rename = "port.redetect_survivor_ratio",
+        skip_serializing_if = "redetect_is_off"
+    )]
+    pub port_redetect_survivor_ratio: f32,
 
     // ── estimator ───────────────────────────────────────────────────────
     /// Which linearization runs.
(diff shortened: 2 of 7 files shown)


pablo-rerun:
Fixed review r2 in e40a1011 (tests and D75 only; runtime behavior and config values unchanged).

Added five public-API frontend tests using the synthetic rig and real CPU tracker/detector:
- `redetection_resumes_only_below_the_survivor_threshold`: controlled losses one track below, at, and above the cutoff.
- `redetection_resumes_on_the_frameset_that_loses_every_track`: total camera-0 loss redetects on the same textured frameset.
- `redetection_uses_the_latest_post_detection_count`: replenishment changes the baseline, and the next decision uses the full post-detection count.
- `redetection_resumes_after_an_empty_initial_detection`: blank startup followed by texture, then another loss/recovery.
- `redetection_keeps_its_baseline_after_a_rejected_frameset`: backend rejection after camera 0 tracking; unchanged retry skips, later loss redetects and matches a clean run.

Temporarily replaced the final `should_detect` comparison with `false`: all five tests failed. Restored the original source before the final checks and commit.

D75 now defines unconnected observations as all observed IDs absent from the landmark database, including carried tracks after failed triangulation or landmark removal. Those tracks can vote on skipped-detection frames, so gating alone does not imply starvation.

Validation through the specified Pixi environment and `/tmp/slam-rs-s32/redetect-target`:
- `cargo test --workspace --release`: 452 passed, 0 failed.
- `cargo clippy --workspace --all-targets -- -D warnings`: passed.
- Touched-file rustfmt check: passed. Workspace `cargo fmt --all -- --check` still fails only on the pre-existing `estimator/mod.rs:1287` formatting difference.

Summary lines:
```text
test result: FAILED. 0 passed; 5 failed; 0 ignored; 0 measured; 40 filtered out; finished in 0.01s
test result: ok. 45 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.11s
    Finished `dev` profile [optimized + debuginfo] target(s) in 1.94s
```

Rollback limit: the baseline assignment follows the last fallible operation. This tests baseline preservation across snapshot/restore, not rollback after an already changed baseline.

No harness run is needed: behavior is unchanged. No `.rrd` was produced.

pablo-rerun:
Ratio changed 0.7 -> 0.85 (d06fc98f). On the 76 s MIO07 the 0.7 gate drifted out of band: ATE vs GT 2.624 cm against 2.29 allowed (LM cap alone 2.093; 0.7 without the cap 2.682, so the gate is the cause). At 0.85 MIO07 reads 2.201 cm at 3.858 ms median (cap-only 4.073). Re-gated on MIO10 against the wave-1 stack: 3.558 -> 3.347 ms, ATE 1.525 -> 1.481 cm, zero lost; MGO09 byte-identical to the stack (the gate never fires on its 107 framesets). D75 amended with the numbers.

PR #257, opened 2026-09-10
slam-rs: replay a reference segment from a catalog server (--catalog)

pablo-rerun:
`tools/apps/replay.py --catalog rerun+http://dgx-spark:9988 --segment <id>` opens the segment by id on the catalog server, ground truth from the same dataset's layer, instead of the manifest's NAS file paths. A laptop or a device with no NAS mount now replays a reference clip from its id alone; before this, that needed two `scp`s and `--rrd/--gt-rrd`. `--catalog` with `--rrd` or `--gt-rrd` is refused as two sources for one replay.

The feed's `CatalogSegment` did this already; the change is the option and the source selection in `replay.main`, three tests that pin which source is opened without a network (the feed is replaced), and the README command. Fast suite 273 passed, ruff, vulture and pyrefly clean. Live run from this host against the catalog: see the comment below.

### DIFF
--- a/packages/slam-rs/README.md
+++ b/packages/slam-rs/README.md
@@ -44,6 +44,7 @@ pixi run -e slam-rs-dev --frozen python tools/apps/replay.py --stage vio   # the
 pixi run -e slam-rs-dev --frozen python tools/apps/replay.py --stage vio --rr-config.headless --rr-config.save data/replay-vio.rrd
 pixi run -e slam-rs-dev --frozen python tools/apps/replay.py --stage vio --segment <segment-id>       # another reference segment
 pixi run -e slam-rs-dev --frozen python tools/apps/replay.py --stage vio --rrd base.rrd --gt-rrd gt.rrd   # a recording of your own
+pixi run -e slam-rs-dev --frozen python tools/apps/replay.py --stage vio --catalog rerun+http://dgx-spark:9988   # from a catalog server, no NAS mount needed
 ```
 
 In a shell without `DISPLAY`, pass `--rr-config.headless` or the spawned viewer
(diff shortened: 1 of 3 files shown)


pablo-rerun:
Live run from pablo-dl-server against the catalog, no NAS path involved:

```
python tools/apps/replay.py --stage vio --gpu --profile fast --catalog rerun+http://dgx-spark:9988 --rr-config.headless --rr-config.save ...
replaying msd-index__MIO_others__MIO10_short_2_panorama (smoke tier) from rerun+http://dgx-spark:9988
2 cameras, 412 framesets, ground truth attached
412 framesets in 6.0 s (68.9 fps)
ATE rmse=1.55 cm  max=4.34 cm  median=1.17 cm   (vs ground truth)
```

Recording: https://pablos-4800gt.ilish-ruler.ts.net:8768/slam-rs/slam-rs-mio10-fast-from-catalog-5c58af35.rrd (headless viewer screenshot checked: trajectory against ground truth, landmarks, both camera panes with keypoints, timing and ATE plots all render).

pablo-rerun:
45c5160a: `--catalog` now takes any segment of a dataset the manifest knows, not only the reference set. Live from pablo-dl-server, the segment that was refused before:

```
--catalog rerun+http://dgx-spark:9988 --segment msd-g2__MGO_others__MGO11_short_3_backandforth --gpu --profile fast
replaying ... (not in the reference set: ground truth only) from rerun+http://dgx-spark:9988
4 cameras, 539 framesets, ground truth attached
539 framesets in 9.3 s (58.1 fps)
ATE rmse=2.30 cm  max=4.73 cm  median=2.14 cm
```

Matches the S32 catalog run's number for this clip (2.30 cm). Datasets known today: msd-index and msd-g2; msd-odyssey has no manifest block yet.

PR #265, opened 2026-09-11
slam-rs: any RoboCap session replays from the catalog

pablo-rerun:
Same fix as #258, for RoboCap. `robocap_probe.py --session sNNNNNNNN` only accepted the one session the manifest lists (s15, the one with a checked-in regression reference); the catalog holds many more sessions of the same device. Now any `s`-plus-eight-digits id resolves to `robocap__<device_id>__<session>` on the catalog; an unlisted session has no regression reference and the probe says so and reports it unscored. A malformed id still errors, naming the shape and the listed sessions.

`RobocapReference` gains `device_id` (already in the manifest) and `is_listed`. 3 files, +31 / -7. ruff, pyrefly, 258 fast tests.

### DIFF
--- a/packages/slam-rs/slam_rs/apis/robocap_probe.py
+++ b/packages/slam-rs/slam_rs/apis/robocap_probe.py
@@ -59,6 +59,8 @@ def main(config: Config) -> None:
     """
     manifest: ReferenceManifest = load_manifest()
     session: RobocapSession = manifest.robocap.session(config.session)
+    if not manifest.robocap.is_listed(config.session):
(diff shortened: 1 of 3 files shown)


pablo-rerun:
Live from pablo-dl-server, an unlisted session (s01) through the catalog, GPU fast, headless, first 15 s:
```
s00000001 is not in the manifest: replaying it from the catalog with no regression reference
329 framesets fed, 329 tracked, 0 refused and retried, 0 lost, 5497 inertial samples pushed
Vio.track 3.4 ms per frameset (median 2.3, max 142.4)
```

PR #266, opened 2026-09-11
slam-rs: pyserde for documents, the manifest becomes gate.toml, typed calibration accessors

pablo-rerun:
Stacked on #265. Follows Pablo's decision that Python (de)serialisation of documents in this repo goes through pyserde, after a three-way sweep (Python sites, Rust sites and the PyO3 boundary, pyserde docs and repo verified against the installed 0.31.8).

## What
- **Dependencies.** `tomli-w` and `orjson` added beside every `pyserde` declaration (`common`, `catalog-common`, `gsplat-rust-renderer`, `slam-rs-osx`). The conda `pyserde` ships no extras, so `from_toml` failed to import before; `orjson` is picked up automatically for JSON. Whole-workspace re-lock: only those two packages added, no other version moved.
- **`reference_segments.toml` -> `gate.toml`**, schema 10. It holds what the catalog cannot tell us (the rigs' IMU models and configs, RoboCap's rig quirks) and our gate (tiers, hold-outs, twenty baseline rows). The copied catalog facts (capture, gt, layer fingerprints, dataset geometry) are gone with their catalog-agreement tests; none had a runtime reader. 533 -> 378 lines.
- **Loading through pyserde.** Seven document classes: `@serde(type_check=coerce, deny_unknown_fields=True)` above the frozen slots dataclasses. The 120-line hand loader, `_one_of` and every cast site are gone; the checks pyserde cannot do (schema version, duplicate ids, duplicate `(profile, lane)` rows, finite positive baselines, dataset existence) stay as a post-load validator, and every error carries the file name. Round-trip and refusal tests added. The config profile overlay is untouched (text-level, Rust-owned format).
- **Typed reports.** `FleetClipReport`/`FleetReport`, `RobocapReport`, `ClipMetadata`; `Machine` serialises as a nested document. JSON keys and order unchanged (the existing key-order test still reads raw JSON). Unscored numbers are `null`, never `NaN`.
- **Typed calibration accessors** on the Rust `Calibration` pyclass (cameras, intrinsics, extrinsics, IMU), declared in `_core.pyi`; the one Python site that parsed `to_json()` output now reads objects.
- **AGENTS.md `## Serialization`** section, and README/design-notes mention the gate file.

25 files, +507 / -713.

## Gate
- `pixi lock --check` clean; Rust fmt, clippy (both feature sets), 402 release tests; core rebuilt.
- ruff, pyrefly, vulture, fast `pytest` (seconds), slow catalog lane.
- Live from the catalog: release tier passes with every ATE equal to the re-seed to nine decimals; MIO10 replay 1.553 cm (pixel check of the recording in the report); RoboCap s01, 10 s, zero lost; the fleet JSON prints the twelve keys in the old order.


pablo-rerun:
Review: P1 (malformed TOML bypassing the file-named error) fixed in 63c05693; P1 (AGENTS.md vs the config overlay) and the P2 (self-comparing IMU assertion) addressed in 47d38ae2 and 6728d340. The `## Serialization` section is rewritten from a sweep of how the monorepo actually uses pyserde (105 `@serde` classes, 33 with array fields; records with arrays are normal, streams are not): records yes / streams no, explicit-dtype jaxtyping, per-schema strictness, validation at the loader, Rust-owned formats with the overlay as the one sanctioned exception.

PR #267, opened 2026-09-11
slam-rs: the Metal lane stops paying 7 ms per read (CubeCL 0.11.0-pre.3, wgpu 30)

pablo-rerun:
## Why
On the M4 Mac mini the portable GPU lane ran MIO10 at 18.8 ms per frameset because every synchronising read cost 7 ms of host time (S32-E6). The cause is located: wgpu-hal 29's Metal `wait` polls the command buffer status with a 1 ms `thread::sleep` loop; wgpu 30 waits on a condition variable (gfx-rs/wgpu#9328). wgpu 30 is only reachable through CubeCL 0.11.0-pre.3, so this PR is that bump plus the API fallout, with no change to kernels' arithmetic, launch shapes or read points.

## What changed
- `cubecl`/`cubecl-wgpu` pinned `=0.11.0-pre.3`, `wgpu` 30 (resolves 30.0.1). Launch args move to slices/`BufferArg`, `SharedMemory<T>` to `Shared<[T]>`, `CubeDim::new_3d`, reader under `future::reader`. Cargo.toml comments re-verified against the 0.11 sources (Metal family check now exists; the exclusive-memory alignment workaround is still needed).
- One kernel control-flow rewrite: CubeCL 0.11 emits C++ lambdas for `while` conditions and Metal rejected the generated cell-selection shader when loaded from the Python process (the Rust test binary compiled it). The two selector loops became `range_stepped` with the same visits and order.
- The no-Vulkan-adapter test is marked inapplicable on macOS (it disables Vulkan ICDs; Metal ignores that).

## Numbers
Mac mini (M4), fast profile, tracker median ms / ATE cm, catalog replay:

| clip | Metal before | Metal after | Mac CPU lane | accuracy |
|---|---|---|---|---|
| MIO10 | 18.69 / 1.552 | 8.10 / 1.552 | 4.98 / 1.553 | in band |
| MIO07 | 18.85 / 1.957 | 8.14 / 1.957 | 5.10 / 2.092 | in band |
| MGO07 | 21.34 / 2.373 | 9.26 / 2.373 | 8.58 / 2.375 | in band |
| MIO14 | not measured | 7.97 / 6.760 | not measured | **6.76 > 6.61 limit (1.12x the 5090 baseline)** |

Seam bench, two cameras: reads 18.4 -> 9.3 ms, whole 20.3 -> 10.3 ms. Isolated empty-read slope 7.05 -> 1.35 ms per read (5.2x, not the 60x hoped for): a native sample shows the remaining fixed cost in CubeCL 0.11's device-channel worker, which idles on a 150 us `nanosleep`.

RTX 5090, Vulkan, A/B harness against main's core: **byte-identical trajectories and states on MIO10/MIO07/MGO07**; latency +12% / +7.6% / +2.5% (MIO10 1.363 -> 1.526 ms), all of it in the stereo stage (0.24 -> 0.37 ms), the one read-bearing stage on that lane.

## Status: draft, gates failed on purpose
Passing: fmt, clippy (both feature sets), 402 workspace tests, strict rustdoc, the locked wgpu GPU task on Linux (450 tests) and on the Mac (449, 2 ignored), 257 Python tests, byte identity on the 5090.
Failing: the 5090 speed clause (two-camera +8-12%), the Mac MIO14 accuracy band (no before-number exists for it on the Mac), and the S36 target (Metal >= 1.5x the Mac CPU lane).
Follow-ups under investigation before this leaves draft: the 150 us channel sleep (likely both the Mac's residual 1.35 ms/read and the 5090 stereo regression), one read per frameset instead of two, device time per kernel on the new compiler stack, MIO14 before-number on the Mac.

Report with every command: `/tmp/fleet-artifacts/slam-rs/cuvslam/reports/s36/s36-1-metal-readback.md` on pablo-dl-server.

### DIFF
--- a/packages/slam-rs/Cargo.lock
+++ b/packages/slam-rs/Cargo.lock
@@ -143,6 +143,8 @@ dependencies = [
  "event-listener-strategy",
  "futures-core",
  "pin-project-lite",
+ "portable-atomic",
+ "portable-atomic-util",
 ]
 
 [[package]]
@@ -156,6 +158,80 @@ name = "autocfg"
 version = "1.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f2032f911046de80f0a198e0901378627c33f59ea0ac00e363d481118bd70a53"
+
+[[package]]
+name = "awint"
+version = "0.19.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b2a8f298efd080b0d3e1ee1723b5d88ecc0d78805fad347e09b76dbd4dadbe37"
+dependencies = [
+ "awint_core",
+ "awint_dag",
+ "awint_ext",
+ "awint_macro_internals",
+ "awint_macros",
+]
+
+[[package]]
+name = "awint_core"
+version = "0.19.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c24d226d9faa090abec192c8ad1ea74635e6cc94000c72a95ceb60cf691428c3"
+dependencies = [
+ "awint_internals",
+ "const_fn",
+]
+
+[[package]]
+name = "awint_dag"
+version = "0.19.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dc5377016f3077b18304da4233a28626ea069047ca84c9bc8c8b81b659f11032"
+dependencies = [
+ "awint_ext",
+ "awint_macro_internals",
+ "awint_macros",
+ "smallvec",
+]
+
+[[package]]
+name = "awint_ext"
+version = "0.19.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "01ff843cdefdd48fbe9bc11292a902b7921ff37433ed4124fa02136f83186f58"
+dependencies = [
+ "awint_core",
+ "const_fn",
+]
+
+[[package]]
+name = "awint_internals"
+version = "0.19.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6209d7873685ac95ee41c4b63fe29bff9b2e1106de65d05b106b2b769c178575"
+dependencies = [
+ "const_fn",
+]
+
+[[package]]
+name = "awint_macro_internals"
+version = "0.19.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6b519a9cd37bf73e44e29b3779ca90246412b56db4af4e96d959f5082bb8bdcc"
+dependencies = [
+ "awint_ext",
+ "proc-macro2",
+ "triple_arena",
+]
+
+[[package]]
+name = "awint_macros"
+version = "0.19.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "518a133c2163e8f021f708d7278af27dd6c1bcee96ea672a2d825b13a6931818"
+dependencies = [
+ "awint_macro_internals",
+]
 
 [[package]]
 name = "backtrace"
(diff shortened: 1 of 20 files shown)


pablo-rerun:
Codex read-only review (gpt-6-astra), no P1. Two P2s, to be fixed on this branch together with the backoff follow-up:
- **P2** `crates/slam-rs/src/gpu/kernels/cell_select.rs:154`: the lambda-in-while fix has no test on the path that failed (Python host + Metal). Cheapest coverage: a macOS GPU pytest that builds `Vio(..., gpu=True)`, feeds synthetic IMU and textured stereo frames until detection runs, and asserts non-empty frontend output; run it in the Metal gate.
- **P2** `crates/slam-rs/Cargo.toml:57`: "cannot both be on for one platform" is not an upstream restriction (CubeCL's `AutoCompiler` supports both; its macOS build script wants `VULKAN_SDK` when `spirv` is on). Reword to "selected per target to avoid the Vulkan SDK on macOS and unused shader backends".
Verified clean: every migrated binding keeps handle, element count and position; `range_stepped` visits the same coordinates in the same order as the old loops, including empty ranges; the Clippy allowance is narrow; the Cargo.toml claims about 0.11 match the registry sources; the bundled SQLite from `cubecl-environment`'s cache feature adds C compilation but no system library.
Full text: `/tmp/fleet-artifacts/slam-rs/reviews/s36-1-codex-review.md` on pablo-dl-server.

pablo-rerun:
Codex read-only review (gpt-6-astra) of the patch-wiring commits `84db9fd1..cff66ea8`: one P1, one P2, both being fixed on this branch.
- **P1** `pixi.toml` `slam-rs-build` / `slam-rs-wgpu-build`: their `inputs` omit the patch file, the prepare tool and the prepared source, so a patch-only change re-runs the prepare task but pixi can reuse the cached `_core.so` with the old code (`depends-on` orders execution; it does not add inputs). Fix: add the prepare inputs and the prepared tree/marker to both build tasks' inputs; verify with a patch-only edit after a warm build.
- **P2** `tests/test_prepare_patched_deps.py`: all five tests use a synthetic crate; nothing exercises the real pinned patch or Cargo's resolution. Fix: a CPU-only smoke test that prepares the pinned archive, asserts the patched `channel.rs` digest, runs locked/offline `cargo metadata` and asserts `cubecl-common` resolves from the local path with no registry source; plus run the patch's own `park_tests` explicitly, since dependency unit tests do not run under the workspace test command.
Verified clean: no lost-wakeup path in the patch (handle installed before clients escape; writes precede the SeqCst publish; recheck after the last queue check; shutdown publishes before waking); spin/yield budgets and the client backoff unchanged; sha256 before extraction, `filter='data'` extraction, a failed apply leaves the previous tree and publishes no marker; lock diff is exactly the one entry; path is manifest-relative; all seven cargo tasks depend on the prepare task and the macOS lane inherits it.
Full text: `/tmp/fleet-artifacts/slam-rs/reviews/s36-6-codex-review.md` on pablo-dl-server.

pablo-rerun:
Both review items addressed in b7c1ff37 + 980906c6:
- **P1** the two build tasks now list the patch file, the prepare tool and the prepared tree among their `inputs`; proven by a warm build, a patch-only edit (rebuild observed), and a no-change rerun (skipped).
- **P2** a CPU-only smoke test prepares the pinned archive, checks the patched file's sha256 against the tested fork commit, and asserts through `cargo metadata --locked --offline` that `cubecl-common` resolves from `target/patch/...` with no registry source; a new `slam-rs-patch-test` task runs the patch's own park tests and `slam-rs-rust-test` depends on it. Both pass on Linux and on the Mac mini.
The reviewer's pyrefly failure was the stale benchmark environment; the current root environment type-checks clean.
Ready to leave draft.

pablo-rerun:
Validator (Codex gpt-6-astra, xhigh) must-fixes addressed in 22fdcaf0..7df8e9fe: preparation serialised with a per-destination process lock and an atomic swap (the 2-process and 8-caller races now pass); the prepared tree is fingerprinted and repaired on mismatch; `slam-rs-wgpu-test` builds the wgpu extension first and the Metal Python test asserts the wgpu backend instead of skipping; per-checkout bootstrap documented next to the `[patch.crates-io]` line. Linux and Mac gates green, including a fresh Mac source copy with no extension and no Cargo target.
Two gates stay red and are pre-existing, not introduced here: the harness's frozen cuVSLAM x1.2 speed limit on MIO07 (paired main fails it too; the branch is byte-identical and marginally faster), and the Mac's MIO14 accuracy band (6.7605 cm on both main and this branch against a limit derived from the 5090's baseline).
Full validation: `/tmp/fleet-artifacts/slam-rs/cuvslam/reports/s36/s36-8-validate-claims.md`; hardening report `s36-9-wiring-hardening.md`.

PR #269, opened 2026-09-11
slam-rs: per-host gate baselines, Mac mini and Spark rows, Metal-lane docs

pablo-rerun:
## What
1. **Per-host baselines.** `baseline_for(lane, profile, host)` prefers the row measured on the running host and falls back to the reference host's row (the 5090). Accuracy and speed both use the chosen row; speed stays gated only when the row's host is the measuring host. Hypothesis tests cover the lookup and the loader's duplicate check (now keyed by lane, profile, host). No change on the 5090.
2. **Rows for the Spark (GB10, Vulkan) and the Mac mini (M4, Metal)**, GPU lane, fast profile, smoke + release tiers (MIO10, MGO09, MIO07, MGO07, MIO14): two passes each, agreeing within 5%, core hash recorded. With the rows in place both hosts pass all five segments with speed gated, including the Mac's MIO14 (6.7605 cm on main before and after this stage; previously judged against the 5090's 6.0135 and falsely red).
3. **Docs.** README fleet table and Metal paragraphs updated to the after-state; design notes gain "S36 — the Metal lane's two sleeps" (causes, the park patch and how it ships, the same-run budget and the validation's corrections, rejected experiments, open levers) and decisions D80–D83; AGENTS.md gains the prepare-task / bare-cargo bootstrap gotcha and the `cargo fetch --locked` note for `slam-rs-patch-test`.

## Numbers (fast profile, tracker median ms, GPU vs same-host CPU)
| host | MIO10 | MIO07 | MGO07 | ratio |
|---|---|---|---|---|
| RTX 5090 (Vulkan) | 2.0 / 5.6 | 2.7 / 5.9 | 3.1 / 9.8 | 2.2–3.2x |
| Spark GB10 (Vulkan, arm64) | 2.9 / 5.4 | 3.1 / 5.4 | 4.8 / 9.0 | 1.8–1.9x |
| Mac mini M4 (Metal) | 4.6 / 4.9 | 4.8 / 5.1 | 6.0 / 8.5 | 1.06–1.42x |

## Gates
271 fast tests, lint, typecheck, deadcode, `pixi lock --check`; both hosts' smoke + release tiers pass with the rows. No Rust or Cargo changes. Codex read-only review running; a known item it will confirm: the Mac row's host name is the network-qualified name and should be normalised to the short host name before merge.

Report: `/tmp/fleet-artifacts/slam-rs/cuvslam/reports/s36/s36-13-wrapup.md` on pablo-dl-server.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -230,6 +230,9 @@ linux-64 (pixi 0.70.x) and move back to a public release once the fix ships.
 
 ## Gotchas
 
+- **slam-rs bare Cargo / rust-analyzer bootstrap** — run `pixi run -e slam-rs-dev --frozen slam-rs-patch-deps` once per fresh checkout (`slam-rs-osx-dev` on macOS). Cargo's path override needs the prepared `target/patch/` tree; the Pixi Cargo tasks prepare it automatically. The bump runbook is beside `[patch.crates-io]` in the package's `Cargo.toml`.
+- **slam-rs offline patch tests need their own cache** — from `packages/slam-rs`, run `pixi run -e slam-rs-dev --frozen cargo fetch --locked --manifest-path target/patch/cubecl-common-0.11.0-pre.3/Cargo.toml` once per Cargo home before `slam-rs-patch-test` (`slam-rs-osx-dev` on macOS). The prepared crate resolves standalone; an offline `test-log` miss means this cache is incomplete.
+
 - **Never use pip** — all dependency management goes through Pixi
 - **`hf download` not `huggingface-cli`** — conda's huggingface_hub provides `hf`, not `huggingface-cli`
 - **gradio from PyPI, not conda** — conda's gradio package has missing transitive deps
--- a/packages/slam-rs/README.md
+++ b/packages/slam-rs/README.md
@@ -247,27 +247,30 @@ takes 63 minutes on the fast profile against 81 on the reference. The ten-clip
 gate's hardest clip, `MIO14_moving_props`, reads 9.72 cm on the reference (D71)
 and 6.37 cm on the fast profile.
 
-The same tip on the fleet, fast profile, tracker median in ms for
-`MIO10` / `MIO07` / `MGO07`, one unpinned pass with decode in the same process:
+The fleet's fast-profile tracker medians below are milliseconds for
+`MIO10` / `MIO07` / `MGO07`. Ratios compare GPU with CPU on the same host.
+The 5090 values are the reference rows in `gate.toml`; GB10 and M4 use the
+median of three matched runs per lane from S36. These are different measurement
+sessions, not a cross-machine timing budget.
 
-| device | backend | fast, ms | fast over reference | trajectory against the 5090 |
+| device | backend | GPU vs CPU fast medians, ms | CPU/GPU | ≥1.2x, accuracy in band |
 |---|---|---|---|---|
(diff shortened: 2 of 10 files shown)


pablo-rerun:
Codex read-only review (gpt-6-astra): one P1, being fixed on this branch.
- **P1** `slam_rs/reference.py:139` + `gate.toml:90`: host matching is an exact string compare on `platform.node()`, and the five Mac rows store the network-qualified `pablos-Mac-mini.attlocal.net`. Off the home network the Mac row stops matching: accuracy falls back to the 5090 row and speed becomes ungated (MIO14 would read 6.7605 cm against the 5090-derived 6.6149 and fail). Fix: the short host name (`platform.node().split(".")[0]`) everywhere — machine reporting, stored row keys, lookup, duplicate detection, speed compare — with tests for short, `.attlocal.net` and another suffix; 5090 and Spark rows unchanged.
Verified: lookup semantics (host row preferred, first lane/profile row fallback, `None`); duplicate triples rejected; 5090 rows and behaviour unchanged; all ten new rows match the evidence to the digit (hash, framesets, ATE, medians; repeat pairs within 5%); docs figures agree with the S36-8/10/11 reports, stale budget and the 2x kernel claim removed, D64's "faster" separated from the 1.2x margin, D80–D83 consistent; bootstrap commands correct; schema 10; `pixi lock --check` passes. Not independently re-run by the reviewer (read-only sandbox): the worker's lint/typecheck/test passes. Note: the Mac rows were measured with a browser GPU process active, the Spark with its background CPU load; both are disclosed in the report.

pablo-rerun:
P1 fixed in ad89e425: short host names everywhere (machine report, lookup, duplicate check, speed gating), the five Mac rows now `pablos-Mac-mini`, tests for the plain name and two domain suffixes, loader rejects a dotted host. 280 fast tests, lint, typecheck, deadcode, lock check pass. No Mac/Spark re-run by decision. Ready for review.

pablo-rerun:
Follow-ups noted from the Pi 5 run on Trixie (Mesa 26.2.1, report `s36-17-pi-trixie.md`), not in this PR:
- **Pi GPU lane is now correct** (22/22 kernel tests, smoke ATE in band, zero GPU resets) but 13–17x slower than the Pi's CPU lane: MIO10 tracker 281 ms vs 21 ms. Per-stage breakdown puts all of it in the two KLT stages (track 242 ms, stereo 51 ms); pyramid and detection are faster on the Pi's GPU than its CPU. The per-keypoint KLT kernel (one 64-thread group per point, serial thread-0 reductions, group barriers, 5 iterations x 4 levels) is latency-bound on a ~12-unit, 16-wide GPU with ~200 points to hide it. Fix = the KLT redesign already identified for the Mac (SIMD-local reductions, several points per group); until then the Pi uses the CPU lane.
- **Start-up guard**: on Mesa 24.2 the lane ran and produced a 54 cm trajectory because the u8/u16 round-trip probe passes while `shaderInt8`/`shaderInt16` are absent. Needed: refuse the GPU lane (typed error, CPU fallback) when those features are missing, plus a small kernel self-test at start-up.
- Docs: Pi row in the portability table; Trixie renamed the ICD to `/usr/share/vulkan/icd.d/broadcom_icd.json` (no arch suffix; the old name yields zero adapters with exit 0).

PR #271, opened 2026-09-15
slam-rs: run catalog VIO with sensor calibration and GPU decoding

pablo-rerun:
Run SLAM on an existing catalog recording and register a replaceable `slam_rs` pose layer. Defaults to the fast profile, with GPU tracking and CUDA decoding when available; preserves the base recording and blueprint.

- Read IMU calibration and applied timestamp corrections from the catalog; add DataForge ingestion and backfill support.
- Separate runtime settings (`slam.toml`) from regression cases (`benchmarks.toml`).
- Fetch camera packets in bulk and reuse decoders across segments.

Validated: 362 tests, lint/type checks, two ground-truth smoke benchmarks, and a complete Cap A run with viewer screenshots. Backfilled 42 RoboCap and 64 MSD sessions without changing original recordings.

Stacked on #270; merge that first, then retarget to `main`.


packages/slam-rs/slam_rs/apis/catalog_layer.py
```diff
@@ -79,9 +74,9 @@ def main(config: Config) -> None:
         backend: str = str(_core.gpu_backend) if vio.gpu else "cpu"
         print(f"{config.segment}: profile={config.profile}, backend={backend}, decode={decode_device}, load={load_s:.1f}s, "
               f"{len(feed.frame_t_ns)} framesets, cameras={feed.camera_positions}", flush=True)
-        run: SegmentRun = _drive(feed, Lockstep(vio), config_sha256=config_text_sha256(config_text))
+        run: SegmentRun = drive(feed, Lockstep(vio), config_sha256=config_text_sha256(config_text))
```
pablovela5620, line 77: what is this drive function? I dont fully understand waht this is doing here

PR #274, opened 2026-09-16
slam-rs: one-command demo and outsider quickstart tooling (HuggingFace data, local catalog, gate)

pablo-rerun:
## What

The tooling an outsider needs to run slam-rs on the Monado SLAM Dataset without our catalog server or NAS. Second of three PRs after #273; the README rewrite and docs pages follow in the third.

- **Data on HuggingFace.** All 64 converted recordings, three layers each (`base`, `gt`, `sensor_metadata`) plus the six blueprints, at [pablovela5620/msd-rrd](https://huggingface.co/datasets/pablovela5620/msd-rrd), mirroring the converter's output layout.
- **`slam-rs-demo`**: the one command for someone who just cloned. It owns only the replay of the four-camera MGO07 clip on screen; `depends-on` chains the build, the sample download, `slam-rs-catalog-up` and `slam-rs-register`, so each piece exists once.
- **`slam-rs-catalog-up`** is a Python entry (`slam_rs.apis.catalog_up`, thin shim), not a shell string: reuse a catalog that answers the catalog API, else spawn `rerun server`, wait until the API answers, and only then record the server's own pid; a failed file-limit raise is reported. Tested against a real server on a free port, including the never-ready failure path.
- **Download tasks** in two tiers, guarded by the recordings themselves across all three layers rather than marker files, so a clip added to a tier is fetched by everyone and an interrupted download is not mistaken for a complete one: `slam-rs-download-sample` (MIO10, MGO09 and the demo's MGO07, about 50 MB) and `slam-rs-download-all` (all 64 recordings, 15.5 GB).
- **`slam-rs-serve`**: the local in-memory `rerun server` on 51235, same shape as `dataforge-serve`.
- **`slam-rs-register`**: runs dataforge's register tool once per headset with `DATAFORGE_OUTPUT_ROOT` pointed at the download, which mirrors dataforge's layer-major layout; dataforge regenerates the blueprints on register, so the download does not fetch them. Linux only, like every dataforge import in slam-rs.
- **`slam-rs-gate`**: `fleet_check` renamed to `gate` (module, shim, tests, docs); the tool's name no longer refers to our machines.
- **Local catalog by default.** `slam.toml` now points at `rerun+http://127.0.0.1:51235`; another catalog is passed with `--catalog`, as the README already shows.

## Quickstart this enables

```bash
git clone https://github.com/rerun-io/examples-monorepo.git
cd examples-monorepo/packages/slam-rs
pixi run -e slam-rs slam-rs-demo
```

- The gate task no longer depends on `slam-rs-build`: pixi's output fingerprint would otherwise rebuild the CPU core over a GPU one before `--gpu` ran.
- The rename is a deliberate break, no compatibility shim: `slam_rs.apis.fleet_check`, `tools/apps/fleet_check.py` and the default `fleet_check.json` report name are gone; nothing in the repo imported them.
- Cleanup pass (four parallel reviewers) and a thermonuclear review folded in: the register loop lives in one task, the marker-file bug is gone, the file-limit failure is no longer silenced, and the tests use the typed benchmarks loader.

## Verification

- Tests written first (`tests/test_register.py`, `tests/test_quickstart.py`), then the implementation: 316 passed; `typecheck` 0 errors; `lint` and `deadcode` clean; `pixi lock --check` unchanged (no new dependencies).
- `slam-rs-demo` run as a fresh user would (twice, before and after the cleanup), in the `slam-rs` environment from `packages/slam-rs` with a display: the environment installed, the core built, the sample downloaded, the catalog started, both headsets registered, and MIO10 replayed in the viewer at 1.55 cm ATE. The gate then passed both smoke clips against their baselines.


packages/slam-rs/slam_rs/apis/register.py
```diff
@@ -0,0 +1,57 @@
...
+from dataclasses import dataclass
+from pathlib import Path
+
+from rerun.catalog import CatalogClient, DatasetEntry, OnDuplicateSegmentLayer
+
+from slam_rs.config import load_slam_config
+
+
+@dataclass(slots=True)
+class Config:
```
pablovela5620: why do we have this file at all? we're in the monorepo right? we can just lean on dataforge? we shouldn't be brining this in?

pablo-rerun: Agreed, and gone in b73e4297. After #271 the slam-rs env already installs dataforge on Linux, and the HuggingFace download mirrors dataforge's layer-major layout, so `slam-rs-register` now just runs `dataforge/tools/apps/register.py msd --device {index,g2,odyssey}` with `DATAFORGE_OUTPUT_ROOT=data/msd-rrd`. The slam-rs register module, its shim and its tests are deleted; dataforge regenerates the blueprints on register, so the download tasks no longer fetch them. One consequence: registration is Linux-only now (dataforge is a linux-only dep), so the macOS bullet in #275 points at `--rrd`/`--gt-rrd` replay instead. Verified on the host: serve, register twice (idempotent), gate smoke passes both clips; tests/typecheck/lint/deadcode clean.

PR #287, opened 2026-09-19
simplecv: batched UmeTrack skinning + exoego schema sections for SHOW3D (hands, objects, captions, blur, provenance)

pablo-rerun:
Groundwork for ingesting SHOW3D (facebook/show3d-dataset) into dataforge. PR 1 of 5; the design record is the shared-understanding brief (https://pablos-4800gt.ilish-ruler.ts.net:8768/show3d/shared-understanding.md) and the scratch-ingest pixel evidence is at https://pablos-4800gt.ilish-ruler.ts.net:8768/show3d/sample-ingest-evidence.html.

**simplecv.umetrack_temp**
- `skin_landmarks`/`_skin_points` claimed `...` batch dims but reshaped the shared model per batch element, so any batch > 1 raised. Now one call skins N poses: the model is broadcast once and the LBS is two einsums (measured 26x faster and 7x less memory than the per-vertex intermediates for 2000 frames x 788 vertices).
- New `skin_mesh` (dense bone weights) and `wrist_for_hand` (the model is a left hand; the mirror lived in five places). `_skin_points` takes the model; the skinning-transform call is keyword-only (two same-shaped args were swappable). Torch backend kept line-for-line parallel; single-frame results unchanged.
- Tests: batch-vs-frame equivalence over several batch shapes and both backends, identity rest pose, right-hand mirroring without input mutation, and a SHOW3D golden frame (checked-in subject profile + one pose record, read through pyserde) reproducing the shipped `landmarks_3d_mm` to < 1e-3 mm.

**exoego_schema.md** (still `exoego:v2`, additive)
- §1-4: device-anchored world (SHOW3D's back rig), multi-camera static exo rigs, `reference` omitted when no sensor sits at a static rig's origin, optional per-frame headset validity flags, `root_T_world` for a later SLAM layer.
- §5/§6: root `AnnotationContext` at `/` (matches every existing writer) with a class-id table (0 = COCO-133, 1 = UmeTrack hand); `frame_index` sequence timeline beside `video_time` for frame-indexed sources.
- New §10-14: hands, objects, captions/text, face-blur boxes, per-frame source provenance. Writers land in the dataforge SHOW3D PRs that follow.

Gate: simplecv-dev gate on pablo-dl-server @ 4e601aac

- Rerun pixel evidence: not applicable (no Rerun view changed in this PR).

### Follow-ups

- §10 rewritten: hands land in the shared COCO-133 stack (`world/gt/coco133_xyz`, `Points3DWithConfidence`, class 0) and the shipped pixels in `<pinhole>/coco133_uv`; the dataset-specific class 1 is gone. Pane rule for rectified versus distorted cameras documented.
- `simplecv/__init__.py`: beartype now checks function boundaries only (`claw_is_pep526=False`), like dataforge, exo-calib and slam-rs. Annotated jaxtyping locals inside `assembly21_to_coco133` rebuilt a checker per call: 3.3 s of a 3.9 s SHOW3D hand layer write, now 0.08 s.


packages/simplecv/simplecv/umetrack_temp/generic_hand_model_numpy.py
```diff
@@ -239,23 +240,23 @@ def hat(v: Float32[ndarray, "n 3"]) -> Float32[ndarray, "n 3 3"]:
 
 
 def _finger_fk(
-    joint_local_xfs: Float32[ndarray, "... dof_per_finger 4 4"],
-    parent_transform: Float32[ndarray, "... 4 4"],
-) -> list[Float32[ndarray, "... 4 4"]]:
+    joint_local_xfs: Float32[ndarray, "batch dof_per_finger 4 4"],
```
pablovela5620: why did you change this to batch like this? instead of ...? can you look at the jaxtyping documentation so we have clarity as to when each should be used?

pablo-rerun: The change is deliberate, and the short version is that `...` was claiming something the code does not do.

In jaxtyping:
- `...` = zero or more anonymous leading axes, never checked (it is `*_`).
- `*batch` = zero or more named leading axes, checked consistent across every argument that uses the name.
- `batch` = exactly one axis, checked consistent across arguments.

`_finger_fk` and `_joint_local_transform` index `[:, i]` and `reshape(shape[0], -1, 4, 4)`, so they only work with exactly one leading axis. With `...` an unbatched `"20 3"` input would silently index the wrong axis and beartype would let it through. `batch` makes that a type error at the boundary.

The public functions (`skin_landmarks`, `skin_mesh`, `_skin_points`) keep `...` because they really do flatten arbitrary leading dims. The more honest hint there is `*batch`, which would also check that `joint_angles` and `wrist_transforms` share their leading dims. I have not switched those yet because `_skin_points` broadcasts `wrist_transforms`, so a caller could legitimately pass a lower-rank wrist; I will confirm no caller does and switch them in a follow-up rather than in this PR.

Rule I am adding to the python-conventions skill: name the axis when the body indexes it, use `*name` when the function truly broadcasts over leading dims, reserve `...` for "any shape, dtype only".

pablo-rerun: Correction to my reply above, now that I have tested it in our setup (simplecv-dev, jaxtyping 0.2.36, beartype 0.22.9):

- Under `beartype_this_package()` each array is checked **on its own**: dtype, rank, literal sizes, and a name repeated inside one string. A name is **not** compared across arguments or with the return value. So `batch` rejects an unbatched `"20 3"` input (wrong rank), but it does not check that two arguments share the same batch size, and `*batch` checks only the trailing dims. Cross-argument checks happen only under `jaxtyping.jaxtyped` / `install_import_hook`, which we are not adopting (the hook also switches off local-variable checks).
- So the names are documentation today, but they should still be correct.

The follow-up is in 0ce4f64a: the internal helpers keep `batch` (they index axis 1); `skin_landmarks`, `skin_mesh` and `_skin_points` now take `*batch` (they flatten and restore any leading dims), and `_skin_points`' wrist takes `*#batch` because it is broadcast. I checked every caller (dataforge `show3d_hands.py`, `test_show3d_objects.py`, `test_umetrack_skinning.py`, `landmarks_from_hand_pose`): all pass matching leading shapes.

